Issue #17, opened 2020-03-20
Create the frontend app base structure using create react app


comoser:
Assumptions:
- TypeScript based
- ESlint configured
- Husky pre-commit hook to lint
- Configured tor nice development env with TypeScript

comoser:
Actually still missing from this PR:

I would like to insert conventional commits in the future commits that we have, so in order to do that I would like to add:
- `commitlint.config.js` file

@s0 I have a question for you. Since I don't have husky configured in the main `package.json` of the repo, when I commit, it totally ignores the git hook, how can we get around this?

s0:
We can actually have packages installed at the root of the project, and commands specified in the root of the project too!

comoser:
Also, on this issue we will enforce `conventional-commits` from now on, on the whole monorepo

comoser:
Pull request with docs updates worth mentioning: https://github.com/reach4help/covidaid/pull/37

PR #18, opened 2020-03-20
[RFC][#17] Setup base frontend structure

comoser:
PR to create the base structure for the frontend application based on a `create-react-app`.

[skip ci]

It's setup for:
- TypeScript
- ESLint
- Husky with pre-commit hook to lint / prettier
- Development env is configured to be dev friendly 🤔 (basically configuring every rule to make sense at the beginning)

[#17](https://github.com/covidaidapp/covidaid/issues/17)


s0:
Looking good, was able to install packages and build no problem! I'll look into the husky stuff now.

s0:
Pushed a commit that gets husky working. pre-commit checks are delegated to each of the workspace projects via specifying a `pre-commit` script.

s0:
I had to remove the `commit-msg` hook as we have no `commitlint.config.js` specified yet. Feel free to create one, and re-add that hook in this PR.

PR #36, opened 2020-03-21
feat(unify-eslint-config): Unify eslint config across monorepo

s0:
As we start introducing new sub-projects and sub-packages, we don't want to have completely independent / vastly-different configs for eslint... this PR unifies them into central "base" configs for js and ts, and then extends those.

I also ran `lint --fix` in `site`

We probably want to do the same thing for the prettier config

### DIFF
--- a/.eslintrc-ts.js
+++ b/.eslintrc-ts.js
@@ -1,0 +1,42 @@
+module.exports = {
+  extends: [
+    '.eslintrc.js',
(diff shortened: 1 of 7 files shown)


luisFilipePT:
@all-contributors please add @s0 for code, doc, ideas, maintenance, review, content and projectManagement.

Issue #55, opened 2020-03-22
Setup redux


jpmarques66:
Following #40, configure redux as the base for data flow in web-client;

PR #68, opened 2020-03-22
[RFC][#55] - Setup redux in web-client

jpmarques66:
**Summary:** This PR will add basic redux setup

**Description:**
In this PR I add redux basic configuration with an example to provide an easier application bootstrap.
If you think the example doesn't make much sense we can delete it. I just think it will come in handy because of all of the typescript code :)


jpmarques66:
> I prefer the implementation of redux using `hooks` instead of the "old way", imo its cleaner and more readable.
> 
> https://react-redux.js.org/api/hooks <- Tell me what you think
> 
> Not so many times we start projects from scratch and have this opportunities, I would advise to add `immer` to the "redux flow" (and not only but out of scope here).
> 
> It reduces a bit the redux boilerplate and makes the code even cleaner and readable.
> Check here -> https://immerjs.github.io/immer/docs/introduction
> 
> NOTE:
> The two suggestions above only apply if:
> 
> * Most of the team feels comfortable with it
> * It integrates fine with typescript (that most of you know I know very little and have been avoiding 😊 )
> 
> Leaving the aprove to @comoser

Thanks for the feedback @luisFilipePT. All good suggestions! I'm going to add them 😄

comoser:
> I prefer the implementation of redux using `hooks` instead of the "old way", imo its cleaner and more readable.
> 
> https://react-redux.js.org/api/hooks <- Tell me what you think
> 
> Not so many times we start projects from scratch and have this opportunities, I would advise to add `immer` to the "redux flow" (and not only but out of scope here).
> 
> It reduces a bit the redux boilerplate and makes the code even cleaner and readable.
> Check here -> https://immerjs.github.io/immer/docs/introduction
> 
> NOTE:
> The two suggestions above only apply if:
> 
> * Most of the team feels comfortable with it
> * It integrates fine with typescript (that most of you know I know very little and have been avoiding 😊 )
> 
> Leaving the aprove to @comoser

These are indeed very good suggestions, feel free to implement them @jpmarques66 !

And thank you @luisFilipePT for the helpful review 👍

web-client/package.json
```diff
@@ -10,9 +10,14 @@
     "@types/node": "^12.0.0",
     "@types/react": "^16.9.0",
     "@types/react-dom": "^16.9.0",
+    "@types/react-redux": "^7.1.7",
+    "immer": "^6.0.2",
+    "prop-types": "^15.7.2",
```
comoser: do we need prop types here?

web-client/src/App.tsx
```diff
@@ -1,27 +1,21 @@
...
+import { Provider } from 'react-redux';
+import { enableAllPlugins } from "immer"
+import ExampleContainer from './containers/exampleContainer';
+import configureStore from './store';
+
+// Later we can check if we need all immer plugins
+enableAllPlugins()
 
 function App(): ReactElement {
+  const store = configureStore()
```
comoser: if ';' is not mandatory, we should add the rule to enforce it, or the other way around. Personally prefer to have it than not

web-client/src/store/index.ts
```diff
@@ -0,0 +1,18 @@
+import { combineReducers, createStore } from "redux";
+import { composeWithDevTools } from "redux-devtools-extension";
```
comoser: imports with double quotes instead of single quotes here

web-client/src/App.tsx
RafaelFerreiraTVD: Personally i also agree with @comoser

web-client/src/store/index.ts
```diff
@@ -0,0 +1,18 @@
...
+const rootReducer = combineReducers({ exampleReducer })
+
+export type AppState = ReturnType<typeof rootReducer>;
+
+export default function configureStore() {
+    // const middlewares = [thunkMiddleware];
+
+    const store = createStore(
+        rootReducer,
+        composeWithDevTools()
```
comoser: we can already leverage the DEV env flag in order to turn this on, not sure on the key, but you can investigate it on the react docs

web-client/package.json
comoser: `interface` and `type` from TS should totally replace the need for `propTypes`

web-client/src/store/index.ts
jpmarques66: @comoser I created an [`issue`](https://github.com/reach4help/reach4help/issues/71) to enforce some rules on eslint. Maybe I'll leave these coding-style related comments for that issue.

web-client/src/App.tsx
jpmarques66: Same as https://github.com/reach4help/reach4help/pull/68#discussion_r396341297

web-client/src/store/index.ts
jpmarques66: Yes, you're right, I'll add that! Thanks

web-client/src/store/index.ts
comoser: makes sense! go for it

web-client/package.json
jpmarques66: > do we need prop types here?

I moved prop-types for devDependencies. Nicely pointed out.

> `interface` and `type` from TS should totally replace the need for `propTypes`

I actually had this doubt as well. From what I read, typescript serves as a check at build time. Prop-types works in runtime. There's actually a utility to infer prop-types from typescript that may become very useful. So, for now, I'd say we keep both. If it becomes evidently useless we'll remove it.

PR #69, opened 2020-03-22
Add CI to build and deploy site using github actions

s0:
gh-pages works well as a CDN, and actions provides good transparency into the build process.

After having thought it over somewhat, and having done this with a number of other production sites, I feel it's best for our landing page too, as it keeps everything in one system (GitHub) and avoids the need to manage users/admins in multiple places.

This deploys the site by pushing it to the `gh-pages` branch of https://github.com/reach4help/reach4help-gh-pages

Example of what a build log looks like: https://github.com/reach4help/reach4help/actions/runs/61003644

### DIFF
--- a/.github/workflows/build-site.yml
+++ b/.github/workflows/build-site.yml
@@ -1,4 +1,4 @@
-name: Build Site
+name: Build & Deploy Site
 
 on: push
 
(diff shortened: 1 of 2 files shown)


sabind:
👍 This will really cut down on CDN hosting costs for our front-ends and make that process seamless :)

.github/workflows/build-site.yml
```diff
@@ -0,0 +1,39 @@
...
+        node-version: 12.x
+
+    - name: Install Packages
+      run: yarn install
+
+    - name: Build Site
+      run: cd site && yarn run build
+      env:
+        SPACE_ID: ${{ secrets.CONTENTFUL_SPACE_ID }}
+        ACCESS_TOKEN: ${{ secrets.CONTENTFUL_ACCESS_TOKEN }}
```
micaelr95: I think the contentful token can be removed because we are no longer using contentful

sabind:
Closing this as we won't be deploying to `gh-pages`.

PR #79, opened 2020-03-23
Initial commit for API project

wr46:
- Initial project base structure (NestJs)
- Docker integration for the database and adminer (postgres)
- Project documentation


api/CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,129 @@
...
+https://www.contributor-covenant.org/version/2/0/code_of_conduct.html.
+
+Community Impact Guidelines were inspired by [Mozilla's code of conduct
+enforcement ladder](https://github.com/mozilla/diversity).
+
+[homepage]: https://www.contributor-covenant.org
+
+For answers to common questions about this code of conduct, see the FAQ at
+https://www.contributor-covenant.org/faq. Translations are available at
+https://www.contributor-covenant.org/translations.
```
comoser: No need for a CoC specific for the API, it's already in the `root`

api/package.json
```diff
@@ -0,0 +1,75 @@
+{
+    "name": "api",
+    "version": "0.0.1",
+    "description": "",
+    "author": "",
+    "private": true,
+    "license": "UNLICENSED",
```
comoser: MIT is the base license

wr46:
> I wouldn’t forget to add phone number to the user records, before we all submit the same PR

This User entity work as an exemple for project structure.

api/src/users/users.service.ts
```diff
@@ -0,0 +1,33 @@
+import { Injectable, Inject } from "@nestjs/common";
+import { Repository } from "typeorm/repository/Repository";
+import { User } from "./entities/user.entity";
+import { CreateUserDto } from "./dto/create-user.dto";
+import { USER_REPOSITORY } from "src/constants";
+
+@Injectable()
+export class UsersService {
```
comoser: suggestion here: you can actually create a generic `repo` which you called `service` and extend from it with the Type inside the <>, like <User>, just to enforce the usage of at least standard methods, like `find`, `findAll`, `findOne`, `create` and `remove`

comoser:
@all-contributors please add @wr46  for code, ideas, "project management", maintenance, documentation

api/src/users/users.service.ts
wr46: This repository must be changed to a simpler structure.

sabind:
#107 adds some missing pieces to the heroku build pipeline.

api/.eslintrc.js
```diff
@@ -0,0 +1,24 @@
+module.exports = {
```
s0: This should extend the base typescript eslint ruleset `../.eslintrc-ts.js` like this: https://github.com/reach4help/reach4help/blob/master/map/.eslintrc.js

api/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+    "module": "commonjs",
+    "declaration": true,
+    "removeComments": true,
+    "emitDecoratorMetadata": true,
+    "experimentalDecorators": true,
+    "target": "es2017",
+    "sourceMap": true,
+    "outDir": "./dist",
+    "baseUrl": "./",
+    "incremental": true
```
s0: Running `start`, `build` or `start:dev` are all giving me typescript errors atm, looks like duplicate definitions for standard libraries.

I think the issue here comes from `styled-components` being installed in some of the other packages in the repo, and polluting the shared `node_modules`, but we'll need this here to be able to compile the api now.

cc @comoser 

Adding `skipLibCheck` seems to fix the issue.

```suggestion
    "incremental": true,
    "skipLibCheck": true
```

s0:
I'm reviewing this atm

api/CONTRIBUTING.md
```diff
@@ -0,0 +1,101 @@
...
+```
+yarn typeorm migration:run
+```
+
+6 - The api swagger-ui is now available under [localhost:3001](http://localhost:3001)
+
+
+# How to report a bug
+
+> If you find a security vulnerability, open an issue describing the problem.
```
s0: We don't want people opening issues to report security vulnerabilities, we want them to disclose ethically. @mdeous will be writing up and publishing our security policy soon, this readme should refer to that when available, until then, change this to the following:

```suggestion
> If you find a security vulnerability, please send a report to <security@reach4help.org>
```

api/CONTRIBUTING.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Code Style Enforced by Prettier
+
+Prettier guarantees the code style adopted and runs on commit, stick to this code style.
+
+<img height="42" width="42" src="https://prettier.io/icon.png" alt="Prettier">[ Prettier site for more information](https://prettier.io/ 'Prettier site for more information')
+
+# PULL Requests
+
+Opening pull requests should be done with enough information and screenshots for visual changes to facilitate the reviewers job. Its MANDATORY to add a link to the issue related
```
s0: Visual changes are not relevant for `api`.

api/CONTRIBUTING.md
```diff
@@ -0,0 +1,101 @@
+# Contributing to web-client for Reach4Help
```
s0: It's `api` not `web-client` :)

api/.eslintrc.js
wr46: We have to make some changes. I did it but then many strange errors appeared.

api/tsconfig.json
comoser: We should be able to specify the `styled-components` dependency only on the web-client, is there a way @s0 ?

s0: With `yarn workspaces`, I don't believe that's possible, unless another package in the workspace is using a different version of that package.

I've used [rush](https://rushjs.io/) in other repos alongside pnpm, which has considerably more advanced dependency management, and prevents phantom dependencies. I prefer it and typically use it on my own projects, the downside however is it's a lot more opinionated, and a little bit more fiddly to add / remove packages, and has a bit more of a learning curve compared to yarn workspaces.

s0: And of course it's yet another development tool to install...

s0: For now, i'd recommend just adding the skip lib check in the projects that need it, and continuing with yarn workspaces.

api/.eslintrc.js
s0: what were the errors?

s0:
I'm happy for it to be merged if @comoser is :)

Issue #77, opened 2020-03-23
Populate map with more organizations (in particular portugal & NYC)


s0:
We need to add more data to our [Mutual Aid Map](https://map.reach4help.org/).

Instructions for adding new data to the map can be found here: https://github.com/reach4help/reach4help/tree/master/map/src/data

If you have information on projects, but are not able or comfortable editing the data files directly, please feel free to enter the information in this form: https://docs.google.com/forms/d/e/1FAIpQLSd4rbN_UPk3RW3a_9sIbBmdv6pH30jqzWKv8Hem9YDRr-uAfQ/viewform

We would love as much help as possible for this, please comment with an overview of the data you will be adding as you start working on this to prevent duplicated/wasted effort.

For a list of sources, see: https://github.com/reach4help/reach4help/wiki/Map-Project:-Unifying-Mutual-Aid-Group-Directories-and-Maps

Added resources so far:
1. (Version: 14:56 - 2020-03-23, A-Kentucky) https://docs.google.com/document/d/1uP49OQGhosfBN4BOYQvyy_Mu3mpCSOYzip13LksC-S8/edit

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/77)

*This information is updated automatically. To modify it, please use ZenHub.*

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/77)

*This information is updated automatically. To modify it, please use ZenHub.*

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/77)

*This information is updated automatically. To modify it, please use ZenHub.*

s0:
I'm currently going through this, and modifying the schema / types as I go as necessary to fit the data :)

s0:
Done (2) all the way up to end of california now, @sabind will be taking over on data entry for a bit.

s0:
Going to work on this again now, starting from Colorado.

s0:
Stopping at Kentucky with #217 today.

kfo42:
https://mutualaid.nyc/groups/ I saw that you needed mutual aid groups from New York for the map. This page has a CSV file containing 107 different mutual aid groups in NYC.

kfo42:
More mutual aid for students:
[...]

PR #83, opened 2020-03-23
Some initial styling, and additional markers

s0:
Added alabama, arizona and california datapoints from here: https://docs.google.com/document/d/1uP49OQGhosfBN4BOYQvyy_Mu3mpCSOYzip13LksC-S8/edit#

Also:
* Changed the data model for markers quite a bit to fit the data I was stuffing in, and make it more flexible
* Added some initial intro paragraphs
* Styled search bar to overlay map and look like other map UI elements
* Update the URL whenever we zoom or PAN, and restore the position form the URL on load / reload (will also enable sharing a specific location)

Ideas I've had, but not for this PR:
* Clicking on markers should zoom to level depending on service radius
* should be able to select a marker from list even when clustered (many markers will share exact same location)
* detailed information should display in sidebar (rather than popup), this will be very important for mobile.
* we shouldn't update the list of filtered results (from viewport) until a user clicks a button to update the results (like google maps), otherwise you can't click through the list of results on the sidebar, because clicking on the first one filters out the rest.

### DIFF
--- a/map/package.json
+++ b/map/package.json
@@ -27,7 +27,8 @@
     "eject": "react-scripts eject",
     "lint": "eslint . --ext .js,.ts,.tsx",
     "pre-commit": "lint-staged",
-    "prettier": "prettier --check \"./**/*.{ts,tsx,js}\""
+    "prettier": "prettier --check \"./**/*.{ts,tsx,js}\"",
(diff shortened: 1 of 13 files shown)


s0:
#77

PR #106, opened 2020-03-23
Adding more data (VT) and fixing bugs

sabind:
Fixed crash with "replaceHistory"

### DIFF
--- a/map/src/components/map.tsx
+++ b/map/src/components/map.tsx
@@ -181,6 +181,53 @@ const debouncedReplaceHistory = debounce((map: google.maps.Map) => {
   );
 }, 299); // max is 100 times in 30 seconds
 
+const haversineDistance = (
+  latLng1: google.maps.LatLng,
+  latLng2: google.maps.LatLng,
+): number => {
+  const lon1 = latLng1.lng();
+  const lon2 = latLng2.lng();
+  const radlat1 = (Math.PI * latLng1.lat()) / 180;
+  const radlat2 = (Math.PI * latLng2.lat()) / 180;
+  const theta = lon1 - lon2;
+  const radtheta = (Math.PI * theta) / 180;
+  let dist =
+    Math.sin(radlat1) * Math.sin(radlat2) +
+    Math.cos(radlat1) * Math.cos(radlat2) * Math.cos(radtheta);
+  dist = Math.acos(dist);
+  dist = (dist * 180) / Math.PI;
+  dist = dist * 60 * 1.1515;
+  dist *= 1609.344; // for meters
+  return dist;
+};
+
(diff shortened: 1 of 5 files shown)


map/src/components/map.tsx
```diff
@@ -165,6 +165,44 @@ function infoWindoContent(info: MarkerInfo): string {
...
+      fn(...args);
+      isCalled = true;
+      setTimeout(() => {
+        isCalled = false;
+      }, wait);
+    }
+  };
+}
+
+const throttledReplaceHistory = throttle((map: google.maps.Map) => {
```
s0: It'd be better to use a debounce rather than a throttle here because we don't want to loose the last call to this function, as that is the one we care about, and may potentially be quite different to a previous call.

Additionally we have `lodash` as a dependency, so we can just use the function defined there:
https://lodash.com/docs/4.17.15#debounce

map/src/components/map.tsx
```diff
@@ -403,9 +430,35 @@ class Map extends React.Component<Props, {}> {
...
 
         // Update labels of markers to be based on index in visibleMarkers
-        m.clustering.visibleMarkers.forEach((marker, index) => {
-          marker.setLabel((index + 1).toString());
-        });
+        m.clustering.visibleMarkers
+          .sort((a, b): number => {
+            const aPosition = a.getPosition();
+            const bPosition = b.getPosition();
+            const mapCenter = map.getCenter();
```
s0: Let's pull this out of the sort function to reduce the amount of calls to `getCenter()`

map/src/data/markers.ts
```diff
@@ -27,6 +27,28 @@ export interface MarkerInfo {
...
+    },
+  },
+  GB: {
+    NEWCASTLE_TYNE: {
+      locationDescription: 'Newcastle upon Tyne, UK',
+      lat: 54.975601,
+      lng: -1.613076,
+      serviceRadius: 9000,
+    },
+    UK: {
```
s0: The UK encompasses GB (it's the United Kingdom of Great Britain and Northern Island), so it probably makes sense to replace `GB` with `UK` here, and then `UK` could become `ALL` perhaps?

sabind: Lol yes. I legit have no idea what's correct. It's been so long since I worked on anything touching that part of the world that I don't remember what it's supposed to be called where 😛.

My recommendation is to have you rename it to whatever is most appropriate.

map/src/components/results.tsx
```diff
@@ -31,5 +31,6 @@ const Results = (props: Props) => {
 
 export default styled(Results)`
   width: 400px;
+  overflow-y: scroll;
```
s0: this doesn't actually fix the issue, but i'll look into that anyway today :)

PR #107, opened 2020-03-24
Database Provider change to support Heroku

sabind:
### Motivation:

Allow Heroku to Build and Deploy the API while connecting to a PSQL DB

### Changes:

- Added `Procfile` that runs `npm run start:prod`
- Added `heroku-postbuild` step to run `nest build`
- Updated `database.provider.ts` to allow either a `DATABASE_URL` (like heroku uses) or the normal dev `.env` file


sabind:
Apologies in advance for stacking all the changes from master back onto this :(

PR #132, opened 2020-03-24
improvement(map-ux): don't zoom on clusters, update right hand list

sabind:
### Motivation: 

Too much zooming makes for bad experience

### Changes:

Clicking a cluster now only updates the right hand list view

Kapture 2020-03-24 at 16 35 44


s0:
Good improvement! :+1: 

I've gone a little further with a couple of related improvements.

PR #148, opened 2020-03-25
ci(map): deploy map automatically when master is updated

s0:
Successful deploy here: https://github.com/reach4help/reach4help/runs/533730215?check_suite_focus=true

### DIFF
--- a/map/package.json
+++ b/map/package.json
@@ -24,7 +24,7 @@
   },
   "scripts": {
     "start": "SKIP_PREFLIGHT_CHECK=true react-scripts start",
-    "build": "react-scripts build",
+    "build": "SKIP_PREFLIGHT_CHECK=true react-scripts build",
     "test": "react-scripts test",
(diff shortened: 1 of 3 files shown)


map/package.json
```diff
@@ -24,7 +24,7 @@
   },
   "scripts": {
     "start": "SKIP_PREFLIGHT_CHECK=true react-scripts start",
-    "build": "react-scripts build",
+    "build": "SKIP_PREFLIGHT_CHECK=true react-scripts build",
```
sabind: What are we actually skipping here? Can we better yet set this flag on the CI vs setting it on the script itself?

sabind:
```suggestion
    "build": "react-scripts build",
```

and in `build-deploy-map.yaml`

```
SKIP_PREFLIGHT_CHECK=true  yarn workspace map build
```

s0:
Addressed, and build working

Issue #150, opened 2020-03-25
Design the ER diagram for the MVP


wr46:
There is a initial diagram on ~~[dbdiagram.io](https://dbdiagram.io/d/5e70e38f4495b02c3b885ca5)~~ we need to design the ER diagram for our MVP.

V2 ERD: [dbdiagram.io](https://dbdiagram.io/d/5e7bc93a4495b02c3b88c6c8)

Action:
- Add the exported PostgreSQL script into ./api/documentation/database/*.sql repo
    - #159 includes a copy of V2 at the moment
- Use [dbdiagram.io](https://dbdiagram.io/home) online tool for the ER design so we can import it after

rgharris:
A few thoughts after reviewing the [workflow](https://www.lucidchart.com/documents/edit/e7b62422-f9ff-4f99-8558-ad99b56e219c/0_0?shared=true) and the schema diagram. Take these with a grain of salt as I am not very familiar with the details of the plans outside of these documents.

1. For `address` in #159 (@sabind), `address2` and `address3` could probably be nullable
2. `request.status` probably should be an enum or a FK? A FK to a status table would be more future proof
3. #127 is considering Auth0 for authentication, if that is the case then `authentication` won't be necessary
4. Will an external service like Twilio Verify be used to verify phone numbers or will that flow happen in the backend? If the latter, will need support for that in the schema.
5. Should `requester_id` and `volunter_id` be named `requester_user_id` and `volunteer_user_id` for clarity?
6. Usually I defer to using `timestamp with time zone` rather than `timestamp`. Here is [one explanation of that](https://justatheory.com/2012/04/postgres-use-timestamptz/)
7. Is the team/organization structure finalized yet? That seems like a pretty specific setup that might be subject to change during/after the MVP. That might could be simplified for now.

Are the following going to be part of the MVP? If so, they may require additional schema. Of course, designing this iteratively and deferring till later rather than upfront is definitely a good approach.

1. Messaging - if so using with something like Twilio Programmable Chat or building internally? 
2. Priority of cases
3. Ratings
4. Tracking offer rejections

puzzledbytheweb:
> Are the following going to be part of the MVP? If so, they may require additional schema. Of course, designing this iteratively and deferring till later rather than upfront is definitely a good approach.
> 
> Messaging - if so using with something like Twilio Programmable Chat or building internally?
> Priority of cases
> Ratings
> Tracking offer rejections

[...]

PR #159, opened 2020-03-25
feat(api): adding start of migrations/entites/dtos from ERD

sabind:
Based on ERD outlined in #150 

Took some liberties with the `address`model and also added a `comment` related to requests. If we like these changes then they should be updated on the ERD


sabind:
@wr46 I'm not familiar with the DTO syntax for relations (Address -> Country for example).

Gave it a stab though :)

api/src/migrations/1585152533866-Country.ts
```diff
@@ -0,0 +1,21 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class Country1585152533866 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+        `CREATE TABLE "country"
+         (
+             id         BIGSERIAL  NOT NULL,
+             name       VARCHAR    NOT NULL,
+             code       VARCHAR(2) NOT NULL,
```
wr46: If we fix the length, it could be CHAR(2). Will not be a varying string.

api/src/migrations/1585152819436-Address.ts
```diff
@@ -0,0 +1,72 @@
...
+               "types" : [ "country", "political" ]
+            },
+            {
+               "long_name" : "94043",
+               "short_name" : "94043",
+               "types" : [ "postal_code" ]
+            }
+         ],
+
+ */
```
wr46: Lets keep this comments until we have a closed model Address. :)

api/src/users/entities/address.entity.ts
```diff
@@ -0,0 +1,62 @@
...
+    this.lng = lng;
+  }
+}
+
+@Entity()
+export class Address extends GenericEntity {
+  @PrimaryGeneratedColumn({ type: 'bigint' })
+  id: number;
+
+  @Column({ name: 'address1' })
```
wr46: We only need the name attribute when entity.attribute don't match with table.attribute.

api/src/users/entities/comment.entity.ts
```diff
@@ -0,0 +1,23 @@
...
+  @PrimaryGeneratedColumn({ type: 'bigint' })
+  id: number;
+
+  @ManyToOne(type => User)
+  @JoinColumn({ name: 'user_id' })
+  requester: User;
+
+  @ManyToOne(type => Request)
+  @JoinColumn({ name: 'request_id' })
+  volunteer: Request;
```
wr46: This would be a User too.

api/src/users/entities/address.entity.ts
sabind: I agree, but I'm worried if someone renames something then we lose the table attribute?

api/src/users/entities/comment.entity.ts
sabind: It's actually supposed to be: 

```
Comment has a user.
Comment belongs to a request.
```

Thank you for catching that though, because it was still very wrong.

PR #172, opened 2020-03-26
Update branding on homepage

s0:
cc @cinthyaejh @r13serra11 

Preview deployed here: https://deploy-preview-172--reach4help.netlify.com/


site/src/components/Helmet.js
```diff
@@ -36,45 +27,28 @@ const Helmet = ({ theme = {} }) => (
...
-          <meta name="image" content={`https:${profile.favicon32.src}`} />
+          <meta name="image" content={logo32} />
           <meta itemProp="name" content={title} />
           <meta itemProp="description" content={description} />
-          <meta itemProp="image" content={`https:${profile.favicon32.src}`} />
+          <meta itemProp="image" content={logo32} />
           <meta name="og:title" content={title} />
           <meta name="og:description" content={description} />
-          <meta name="og:image" content={`https:${profile.bigIcon.src}`} />
+          <meta name="og:image" content={logo512} />
```
luisFilipePT: Please check if this image is working 🤔 

https://www.heymeta.com/url/deploy-preview-172--reach4help.netlify.com/

site/src/components/Header.js
```diff
@@ -50,7 +58,8 @@ const Header = () => (
 
             const homeLink = home && (
               <Image
-                src={Logo}
+                className="logo"
+                src={logo}
                 width="50px"
                 alt="Portfolio Logo"
```
luisFilipePT: "alt="Portfolio Logo" 🤔

s0: good catch! :-D

site/manifest-config.js
```diff
@@ -1,11 +1,32 @@
...
-  name: 'Mate Gatsby Starter',
-  short_name: 'Mate Gatsby Starter',
+  name: 'Reach4Help',
+  short_name: 'Reach4Help',
   start_url: '/',
   background_color: colors.background,
   theme_color: colors.primary,
   display: 'minimal-ui',
-  icon: 'media/icon.png',
+  icons: [
```
luisFilipePT: If I am not mistaken out of the box this already generates all icon sizes, unless you need to override it that I think it's not the case.

We just need the bigger icon and gatsby manifest takes care of the rest
https://www.gatsbyjs.org/packages/gatsby-plugin-manifest/

image

site/src/sections/Landing.js
```diff
@@ -60,41 +61,40 @@ const LandingPage = () => (
...
+              textAlign="center"
+              style={centerHorizontally}
+            >
+              {description}
+            </Heading>
 
             <Flex alignItems="center" justifyContent="center" flexWrap="wrap">
               {socialLinks.map(({ id, ...rest }) => (
-                <Box mx={3} fontSize={[5, 6, 6]} key={id}>
+                <Box mx={3} fontSize={[5, 5, 5]} key={id}>
```
luisFilipePT: If it's the same size for all breakpoints 😄 
```suggestion
                <Box mx={3} fontSize={5} key={id}>
```

luisFilipePT:
Just one more thing, usually `images` folder is inside `src` and not `components`, the rational is that images are available inside `src`. ex: inside `pages` or `templates`. Not an issue just an opinion

site/manifest-config.js
s0: Oh that's great, yep that works!

site/src/components/Helmet.js
s0: Looks like it's working now :)

s0:
@luisFilipePT addressed comments

site/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "build": "gatsby clean && gatsby build",
-    "develop": "gatsby clean && gatsby develop",
+    "develop": "gatsby clean && GATSBY_SITE_URL=http://localhost:8000 gatsby develop",
```
luisFilipePT: `GATSBY_SITE_URL=http://localhost:8000` why?

s0: this is to produce correct URLs for icons when run in development, we need to know the full (public) URL to produce the urls for images correctly (as twitter doesn't support site relative), so i've used it everywhere we generate these URLs.

luisFilipePT: Ok, just asking because I run the app on `8001` because `8000` is already in use by default for my main project 😄 

Thanks for explaining 👍

PR #180, opened 2020-03-26
docs(map): add instructions for how to add markers to map

sabind:
Description
-----------

Add documentation to how markers should be added.

Motivation
----------

There are too many data points to add so we're crowdsourcing this, and thus adding documentation for it.

Testing Guidelines
------------------

Deferring this to [the map README](https://github.com/reach4help/reach4help/blob/master/map/README.md)

Release Checklist
-----------------

- [x] This code has unit tests

Security Checklist
-------------------

If you checked any of those, please request a review from `@application-security`.

Additional Notes
----------------


map/src/data/DOCUMENTATION.md
```diff
@@ -0,0 +1,110 @@
+# Map Marker Documentation
```
s0: I would call this file `README.md` just so it appears in the directory view of `map/src/data` :)

map/src/data/DOCUMENTATION.md
```diff
@@ -0,0 +1,110 @@
...
+
+The purpose of this readme is to outline the data model that the map uses to track organization/community map markers.
+
+## What is a Map Marker
+
+A map marker represents an organization or community effort to orchestrate aid. It consists of several components as well as a latitude, longitude, and area of influence (a radius measured in meters).
+
+### Data Model
+
+```typescript
```
s0: is it worth duplicating this here? perhaps we just add a bit more tsdoc and then link to the relevant lines in the ts? Copying this here leaves potential for it becoming outdated.

sabind: Yes 😞 

I think we can open an issue to run ts doc and to replace this. Once hosting is up we can actually just host all the docs for the map... on the map!

sabind: I'll merge this and PR again to rename.

s0: Hosting is up, we could add this now, though to be honest if someone's going to be editing the same file anyway, i don't think that generating docs from the tsdocs and hosting them anywhere really adds anything?

map/src/data/markers.ts
```diff
@@ -1,29 +1,46 @@
...
+  serviceRadius: number; // Measured in Meters (per Google Maps standard)
 }
 
+/**
+ * A marker that will be rendered on the map. A short title and description is also visible to users.
+ *
+ * It contains an array of services
+ */
 export interface MarkerInfo {
+  // name of the organization or community effort
```
s0: you probably want to use TSDoc syntax `/**` so that these hints are actually rendered for references to these types, but this is good enough for now.

sabind: /** has to be done at the top of the file ?

PR #189, opened 2020-03-27
docs: improve pr template

s0:
Description
-----------

Make instructional comments html comments so they are hidden, and change the mention of application-security to the full handle that should be used.

Motivation
----------

looks cleaner :)


micaelr95:
LGTM

PR #190, opened 2020-03-27
feat(site): add a security.txt well-known file

### LINKED ISSUE #50: Add a security.txt well-known file
Ala: https://securitytxt.org/

We have an email setup: <security@reach4help.org> So we should list it in this file so that security researchers can reach out if necessary.

cc @reach4help/application-security

mdmfernandes:
Description
-----------
Add a security.txt file to the .well-known directory, as per [security.txt](https://securitytxt.org/).

Motivation
----------
Allow security researchers to properly report findings related to this project.

Additional Notes
----------------
closes #50

### DIFF
--- a/site/static/.well_known/security.txt
+++ b/site/static/.well_known/security.txt
@@ -1,0 +1,8 @@
+# If you would like to report a security issue, please use the contact information below
+Contact: mailto:security@reach4help.org
+
+# Our security policy
+Policy: https://github.com/reach4help/reach4help/blob/master/SECURITY.md
+
+# You can contribute to our project, it's open-source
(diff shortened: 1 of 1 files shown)


s0:
This doesn't look like it's actually deploying: https://reach4help.org/.well_known/security.txt

(perhaps gatsby doesn't copy hidden files in static by default?)

Also it should be `.well-known` not `.well_known`

s0:
Uhh nope, looks like it's a problem with netlify: https://community.netlify.com/t/hidden-files-removed-in-zip-deploy/8997

We could add a redirect rule in cloudflare and change the path? @mdeous

PR #195, opened 2020-03-27
Fix issues getting started

kurtharriger:
Description
-----------
Updated README.  env file must be created first. DATABASE_URL needs set (perhaps consider using env-cmd or similar) 

Screen Shot 2020-03-27 at 2 18 37 PM

Added .nvmrc file with latest node version

Screen Shot 2020-03-27 at 2 32 43 PM

Fixed lint-staged warning
Screen Shot 2020-03-27 at 2 13 25 PM

Fixed import error in e2e tests (I've always used relative paths, but it seems that moduleNameMapper will also work if you prefer to use src/) 

Screen Shot 2020-03-27 at 3 26 53 PM

The test itself is still not working, but I do not yet know much about nestjs.

Motivation
----------
Contributors can onboard following readme

Testing Guidelines
------------------
NA

Release Checklist
-----------------
NA

Security Checklist
-------------------

Additional Notes
----------------


api/test/app.e2e-spec.ts
```diff
@@ -6,13 +6,14 @@ import { AppModule } from './../src/app.module';
 describe('AppController (e2e)', () => {
   let app: INestApplication;
 
-  beforeEach(async () => {
+  beforeEach(async (done) => {
```
kurtharriger: Maybe done is not required. I think done was required as workaround for some jasmine issue at one point, but I think its been fixed.  I noticed the error I had was actually failed to start due to not having database_url set first time I ran it.

api/README.md
```diff
@@ -10,14 +10,30 @@ $ yarn install
...
+
+```bash
+# -d to run in background (optional)
+docker-compose up -d
+```
+
+Run migrations
 
 ```bash
+yarn typeorm migration:run
```
wr46: Migration must run after `yarn run start:dev`

api/README.md
```diff
@@ -10,14 +10,30 @@ $ yarn install
...
+```
+
+Start api server [localhost:3001](http://localhost:3001)
```
wr46: Could be better: Access swagger-ui at localhost:3001/api

api/test/app.e2e-spec.ts
sabind: 👍  I think they detect when you return a promise.

wr46: I must push the fix as soon as possible :( for now we can add a require(..) as a temporary fix.

api/README.md
```diff
@@ -10,14 +10,30 @@ $ yarn install
...
+
+```bash
+export DATABASE_URL=postgresql://reach4help_api:reach4help@localhost:15432/reach4help_api_db
```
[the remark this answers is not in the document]
kurtharriger: done

api/example.env
```diff
@@ -1,5 +1,7 @@
 API_PORT=3001
 
+DATABASE_URL=postgresql://reach4help_api:reach4help@localhost:15432/reach4help_api_db
```
wr46: This env var is used only on Heroku. For local dev this will be undifined and TYPEORM_* will be used instead. You can check */database/ormconfig.ts

kurtharriger: hmm I needed it initially, but now it seems to work without it. I think adding dotenv.config() must have fixed it

PR #206, opened 2020-03-29
Only allow requests originating from Cloudflare

### LINKED ISSUE #198: Filter requests not originating from Cloudflare
Now that Cloudflare is setup, we need to ensure the protection it provides can't be bypassed.

For this, the backend needs to check (before any other request processing happens) if the request originates from one of Cloudflare IPs ([IPv4 ranges](https://www.cloudflare.com/ips-v4), [IPv6 ranges](https://www.cloudflare.com/ips-v6)), and drop it if it doesn't.

This is a temporary measure while we are on Heroku, once we move to Azure we will be able to setup a more robust solution.

Relates to #186

mdeous:
Description
-----------
This PR implements an IP filtering system to block any request that doesn't originate from a Cloudflare IP. The feature can be disabled when developing by setting `ENVIRONMENT` to `dev` in the `.env` file.

Disclaimer: I'm not a nodejs dev at all, and this is my first contact with nestjs, be indulgent please 😅. This solution feels a bit overkill compared to the goal, but from what I understood from the nestjs documentation this seems to be the way to go. If you have any idea of how I could make this feature simpler, please leave a comment, I'll be happy to make changes. 

Motivation
----------
To fully benefit from Coudflare's protection we need to ensure all the requests we handle have been proxied through them.

Testing Guidelines
------------------
Locally:
* in the `.env` file, set the `ENVIRONMENT` variable to anything else than `dev`
* make a request to the API
* you should get a 403 error

Release Checklist
-----------------

Security Checklist
-------------------

Additional Notes
----------------
closes #198


api/src/middleware/cloudflare-filter/cloudflare-filter.middleware.ts
```diff
@@ -0,0 +1,29 @@
...
+import { Request, Response } from 'express';
+import { CloudflareIpService } from '../../services/cloudflareip/cloudflare-ip.service';
+import { inRange } from 'range_check';
+
+@Injectable()
+export class CloudflareFilterMiddleware implements NestMiddleware {
+  constructor(private cloudflareIpService: CloudflareIpService) {}
+
+  use(req: Request, res: Response, next: (any?) => void) {
+    if (process.env.ENVIRONMENT !== 'dev') {
```
sabind: `NODE_ENV=development` vs. `NODE_ENV=production` is the standard so I'd keep your code the same, but simply rename `ENV` to `NODE_ENV`

api/example.env
```diff
@@ -1,3 +1,5 @@
+ENVIRONMENT=dev
```
sabind: `NODE_ENV=development`

`NODE_ENV=production`

Should be the two options.

api/src/middleware/cloudflare-filter/cloudflare-filter.middleware.ts
```diff
@@ -0,0 +1,29 @@
...
+      const ipType = req.ip.match(':') ? 6 : 4;
+      this.cloudflareIpService.getIPs(ipType).toPromise().then(
```
kurtharriger: IMO its not necessary to query this on every request as it will introduce additional latency and potential for unexpected failures.

This list is essentially static and given most users use iptables any changes to this list will probably occur way in advance of actually being used. Perhaps we lookup this list at startup or build time and cache it.

sabind:
Closing this since we've removed the API from the project.

PR #229, opened 2020-03-31
improvement(functions): make independent of monorepo

s0:
Description
-----------

~~Somewhat hacky fix to work-around requirement for functions to have an engine defined, by only defining it when we deploy and then removing it...~~

~~Let's not merge this yet, but thoughts???~~

Updated to just move `functions` out of yarn workspaces.

Motivation
----------

Testing Guidelines
------------------

Release Checklist
-----------------

Security Checklist
-------------------

Additional Notes
----------------


sabind:
Looks good, but right now we have a problem since functions is a workspace it's causing issues with yarn install... so maybe we do merge?

s0:
Hmm, i'll briefly discuss it in the next meeting...

s0:
Updated this instead to not be part of the yarn workspaces and have its own lock file, which seems to be the most sensible way of moving forward now tbh.

sabind:
@s0 This was failing initially so I tweaked a little bit the commands inside `functions`. Now `yarn run firebase deploy -P reach4help-<dev|staging|prod>` should work. Tested with `dev`.

s0:
@sabind could you share the errors you were seeing before switching from yarn to npm in the commands? Using both npm and yarn throughout the repo makes me a little uncomfortable, and I'd rather be consistent, so I'm curious what the error was?

sabind:
> @sabind could you share the errors you were seeing before switching from yarn to npm in the commands? Using both npm and yarn throughout the repo makes me a little uncomfortable, and I'd rather be consistent, so I'm curious what the error was?

```sh
$ projects/reach4help: yarn install

yarn install v1.22.4
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
error functions@0.0.1: The engine "node" is incompatible with this module. Expected version "10". Got "12.16.1"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

Unless I do `nvm use 10` prior to working inside the `functions` directory. That being said it does make sense to make it all `v10.x.x` inside there. Idk what's easier honestly.

s0:
Oh right totally, i get what you mean. I think in this case we can take advantage of the yarn `--ignore-engines` flag, and still use yarn throughout.

I pushed an alternative fix here: https://github.com/reach4help/reach4help/tree/firebase/node-engine-2, could you try it out ant let me know what you think?

> It does feel confusing though because `yarn` shouldn't leave little `yarn.lock` files in the sub folders so I don't know if it sets bad expectations and people start yarn installing inside the sub projects?

Running `yarn install` inside of sub projects will work fine, and update the root `yarn.lock` as you'd hope, the only reason that there's a `yarn.lock` inside `functions` is because `functions` is not listed in the `workspaces` property of the root `package.json`. You can safely run `yarn add` `yarn install` etc... anywhere and it won't drop any additional `yarn.lock` files anywhere, except in `functions` :)

> If we instead let yarn call `npm` inside `functions` then we can keep both the `engines` declaration as well as `yarn` as the full orchestrator for all the sub-projects. Those who work on functions can `npm install` as needed, but I agree it adds a layer of complexity that can easily be misunderstood and have the inverse effect where people are calling `npm install` in sub-projects and leaving little `package-lock.json`s everywhere :(

Yeah, i forsee people incorrectly using npm install in other places if we start to do this, hopefully my proposal is a good middle ground? :)

puzzledbytheweb:
https://github.com/reach4help/reach4help/pull/229#issuecomment-606926769

I tried your branch and everything installs properly:
I'm using Ubuntu and node 12.16.1

sabind:
> Oh right totally, i get what you mean. I think in this case we can take advantage of the yarn --ignore-engines flag, and still use yarn throughout.

Sounds good, all we need then is to use --ignore-engines when we use yarn in `firebase.json` as well.

```sh
      "yarn --ignore-engines --cwd functions install",
      "yarn --ignore-engines --cwd functions lint",
      "yarn --ignore-engines --cwd functions build"
```

For example: `yarn --ignore-engines run deploy` or `yarn --ignore-engines install` inside `/functions` both work.

sabind:
I'd say just proceed with either branch as you need, just double test the `pre-commit` hooks because outside of these two branches they do fail which I imagine is annoying to others.

s0:
> Sounds good, all we need then is to use --ignore-engines when we use yarn in `firebase.json` as well.

Huh, weird, looks like i forgot to actually push to the branch i mentioned last night... i made exactly that change! :joy:

This is ready to approve again now.

PR #237, opened 2020-04-01
feat(backend): added models as well as some empty cloud functions

sabind:
Description
-----------

Added a stab at the Data Model.

Remaining work: 

- [x] Add validation that deletes records that don't match the schema when they get created (in case front end validation fails.
- [x] Add unit test framework
- [x] Add Converter functions (still need to add to more objects): https://firebase.google.com/docs/reference/js/firebase.firestore.FirestoreDataConverter

Motivation
----------

Make sure the front end team can proceed as fast as possible, including having access to some entities they can use and store in Firebase.

Testing Guidelines
------------------

I plan on adding unit tests, but right now this code doesn't really do much and likely may change quickly with the data model tweaks.

Release Checklist
-----------------

Security Checklist
-------------------

I am **not** and expert so please correct me if it's all wrong.

I've tried to dissociate all PII from any data that could become PII if attached to a user. As a result the only field that remains is "Username" for the front end to use. Arguably "Display Name" might not be PII is users are properly educated and knowingly publicly disclose their name as part of the ToS + Privacy Policy. If that becomes the case we can add "Display Name" similar to Twitter/IG that allows people to replace their `username` with something more *friendly*

Additional Notes
----------------

Idk if we have issues for the back-end yet. I'm sure this will kickstart that effort.


functions/src/requests/offers/index.ts
```diff
@@ -0,0 +1,159 @@
...
+
+// eslint-disable-next-line no-unused-vars,@typescript-eslint/no-unused-vars
+const queueStatusUpdateTriggers = (change: Change<DocumentSnapshot>, context: EventContext): Promise<void[]> => {
+  const offerBefore = change.before ? change.before.data() as Offer : null;
+  const offerAfter = change.after.data() ? change.after.data() as Offer : null;
+
+  const operations: Promise<void>[] = [];
+
+  // A request has just been accepted -- Update request with new information (CAV, new status)
+  if (offerBefore && offerBefore.status !== OfferStatus.accepted && offerAfter && offerAfter.status === OfferStatus.accepted) {
```
s0, line 112:
```suggestion
  if (offerBefore?.status !== OfferStatus.accepted && offerAfter?.status === OfferStatus.accepted) {
```

functions/src/requests/index.ts
```diff
@@ -0,0 +1,242 @@
...
+
+// eslint-disable-next-line no-unused-vars,@typescript-eslint/no-unused-vars
+const queueStatusUpdateTriggers = (change: Change<DocumentSnapshot>, context: EventContext): Promise<void[]> => {
+  const requestBefore = change.before ? change.before.data() as Request : null;
+  const requestAfter = change.after.data() ? change.after.data() as Request : null;
+
+  const operations: Promise<void>[] = [];
+
+  // A request has just been completed - Update CAV request completed count
+  if (requestBefore && requestBefore.status !== RequestStatus.completed && requestAfter && requestAfter.status === RequestStatus.completed) {
```
s0, line 175:
```suggestion
  if (requestBefore?.status !== RequestStatus.completed && requestAfter?.status === RequestStatus.completed) {
```

functions/src/requests/index.ts
```diff
@@ -0,0 +1,165 @@
...
+  const requestAfter = change.after.data() ? change.after.data() as Request : null;
+
+  const operations: Promise<void>[] = [];
+
+  // We have a new rating -  Update CAV rating average but only this time.
+  if (requestBefore && requestAfter && requestBefore.rating === null && requestAfter.rating !== null) {
+    // TODO: Adjust the avg rating based on the new rating
+    operations.push(Promise.resolve());
+  } else if (requestBefore && requestAfter && requestBefore.rating !== null && requestAfter.rating !== null) {
+    // TODO: Adjust the avg rating based on the old rating and the new rating
```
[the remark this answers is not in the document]
s0: (we could also possibly allow for reviewers to adjust their rating withing e.g. 5mins of giving the rating? we probably want a rating timestamp too...)

functions/src/requests/index.ts
```diff
@@ -0,0 +1,242 @@
...
+
+export interface IRequest extends FirebaseFirestore.DocumentData {
+  cavUserRef: FirebaseFirestore.DocumentReference<IUser> | null;
+  pinUserRef: FirebaseFirestore.DocumentReference<IUser>;
+  pinUserSnapshot: IUser;
+  title: string;
+  description: string;
+  latLng: GeoPoint;
+  status: string;
+  rating: number | null;
```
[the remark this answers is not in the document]
s0: Additionally, we almost certainly want to keep timestamps of rating... in particular if we want to allow for rating adjustments within a certain timeframe of having given a rating? That can probably be v2 tbh...

functions/src/requests/index.ts
```diff
@@ -0,0 +1,242 @@
...
+  const requestBefore = change.before ? change.before.data() as Request : null;
+  const requestAfter = change.after.data() ? change.after.data() as Request : null;
+
+  const operations: Promise<void>[] = [];
+
+  // We have a new rating -  Update CAV rating average but only this time.
+  if (requestBefore && requestAfter && requestBefore.rating === null && requestAfter.rating !== null) {
+    // TODO: Adjust the avg rating based on the new rating
+    operations.push(Promise.resolve());
+  } else if (requestBefore && requestAfter && requestBefore.rating !== null && requestAfter.rating !== null) {
```
s0, line 200:
```suggestion
  if (requestBefore?.rating === null && requestAfter?.rating !== null) {
    // TODO: Adjust the avg rating based on the new rating
    operations.push(Promise.resolve());
  } else if (requestBefore?.rating !== null && requestAfter?.rating !== null) {
```

functions/src/requests/index.ts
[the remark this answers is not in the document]
sabind: Sounds good! Will add

functions/tests/index.test.ts
```diff
@@ -0,0 +1,24 @@
+import { validate } from 'class-validator';
```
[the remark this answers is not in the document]
sabind: Thanks!

functions/src/requests/index.ts
```diff
@@ -0,0 +1,242 @@
...
+  ongoing = 'ongoing',
+  completed = 'completed',
+  cancelled = 'cancelled',
+  removed = 'removed'
+}
+
+export interface IRequest extends FirebaseFirestore.DocumentData {
+  cavUserRef: FirebaseFirestore.DocumentReference<IUser> | null;
+  pinUserRef: FirebaseFirestore.DocumentReference<IUser>;
+  pinUserSnapshot: IUser;
```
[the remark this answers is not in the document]
sabind: > We want PINs to fill in a questionnaire about themselves (e.g. age, other things that make them vulnerable etc...)

I don't think that's an issue. I see a world where PIN/CAV is a workflow, not necessarily a user distinction. 

I might be a nurse helping elderly in their homes, but I will need someone to walk my dogs.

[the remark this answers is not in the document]
sabind: > In the future, not every PIN may have an associated IUser (if a request is being filed on behalf of them by someone else).

*As User I can give someone access to file PIN requests on my behalf.*

Again the goal here is to make sure that we all access the system in a uniform way.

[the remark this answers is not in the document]
sabind: > Clearer separation of auth (and other IUser / PII) data from subject data (I have a feeling this will be useful to us later).

So the auth data goes into a separate database (including all PII). The `users` collection should have no PII. Even if we store address we tie it to the user record which includes no PII so it's segregated.

functions/src/requests/offers/index.ts
```diff
@@ -0,0 +1,159 @@
...
+
+import { IUser, User } from '../../users';
+
+export enum OfferStatus {
+  pending = 'pending',
+  accepted = 'accepted',
+  rejected = 'rejected',
+}
+
+export interface IOffer extends FirebaseFirestore.DocumentData {
```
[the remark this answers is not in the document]
sabind: Will add. Good call. There's some complexity there in terms of write limits to collections with timestamp indexes (https://firebase.google.com/docs/firestore/solutions/shard-timestamp)

I'll incorporate this.

functions/src/requests/index.ts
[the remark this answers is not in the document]
sabind: +1 on rating timestamps.

I disagree that we need two-sided ratings. How I am as a helper and how I am as a PIN (considering I can be both) should maybe fall under one rating: *How I am as a person.*

Knowing that the PIN/CAV I'm interacting with has both helped people and received help gives me confidence they know how to treat others.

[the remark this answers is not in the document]
s0: I agree that we should only show a single rating for "how i am as a person", but we still need to store both:

* what the PIN rated the CAV for this request
* what the CAV rated the PIN for this request

Right now, `rating` is ambiguous, is it the rating of the CAV or the PIN involved in this request? And whatever your role in a request, the rating you receive will affect your personal rating in the same way.

functions/src/requests/offers/index.ts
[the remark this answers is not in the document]
s0: Okay that's good to know,

we can probably forego the any indexes for now until we realise we need them in any queries.

functions/src/models/organizations/index.ts
```diff
@@ -0,0 +1,63 @@
+import { FirestoreDataConverter } from '@google-cloud/firestore';
+import { IsArray, IsNotEmpty, IsString } from 'class-validator';
+import { firestore } from 'firebase-admin';
+import DocumentData = firestore.DocumentData;
+
+export enum OrganizationType {
+  healthCare = 'health care',
+  socialServices = 'social services',
```
s0, line 8: i think for enum values, it's better to not have spaces. perhaps use underscores instead?

```suggestion
  healthCare = 'health_care',
  socialServices = 'social_services',
```

functions/src/models/organizations/index.ts
```diff
@@ -0,0 +1,63 @@
...
+  welfare = 'welfare',
```
s0, line 9: Let's add an organization type for mutual-aid groups?

```suggestion
  welfare = 'welfare',
  mutualAid = 'mutualAid',
```

s0: Also, i'm not really sure of the distinction between these types, are we sure these are the ones we want? Are they clearly defined anywhere?

functions/src/models/organizations/teams.ts
```diff
@@ -0,0 +1,59 @@
+import { FirestoreDataConverter } from '@google-cloud/firestore';
+import { IsArray, IsNotEmpty, IsString } from 'class-validator';
+import { firestore } from 'firebase-admin';
+
+import { OrganizationType } from './index';
+import DocumentData = firestore.DocumentData;
+
+export interface ITeam extends DocumentData {
+  name: string;
+  types: OrganizationType[];
```
s0, line 10: I don't think we want to work on this part of the model yet, it takes focus away from the MVP, and we can add it later. I think we also want to think about this a little more. E.g. what's the difference between a Team and Organization, can't they just be the same thing? etc...

Let's remove this for now.

functions/src/models/questionnaires/index.ts
```diff
@@ -0,0 +1,108 @@
...
+
+export enum QuestionnaireType {
+  pin = 'pin',
+  cav = 'cav',
+  org = 'org',
+  team = 'team'
+}
+
+export interface IQuestionnaire extends DocumentData {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
s0, line 19: this ignore line can go

```suggestion
```

s0:
Acually, changed my mind, i'll wait for you to address the ramining comments.

functions/src/models/users/index.ts
```diff
@@ -0,0 +1,147 @@
...
+
+export interface IUser extends DocumentData {
+  cavQuestionnaireRef: DocumentReference<IQuestionnaire> | null;
+  pinQuestionnaireRef: DocumentReference<IQuestionnaire> | null;
+  averageRating: number;
+  casesCompleted: number;
+  requestsMade: number;
+  username: string;
+  displayName: string | null;
+}
```
s0, line 18: I just realised that we currently have no way of linking firebase auth objects with IUser objects right? We need some way of linking "Firebase ID tokens" to "IUser" objects, correct?

So we probably need to add a uid field to IUser, right? (in the future we probably want it to be optional, to represent people that haven't logged-in, but that can be later). Perhaps making this a `string[]` in case we need to attach multiple auth objects to a single person?

s0: ahh... adding this to the `User` object then makes it public data, and I imagine we don't want to share people's tokens with one another (even if they can't be spoofed).

Perhaps the IUser object needs to be private by default, and have a public sub-collection? something like "PublicProfile"?

s0:
Note to self, given that there are 2 different libraries for `firebase` and `firebase-admin`, we can't use the types directly in a shared library. Luckily there is https://github.com/kossnocorp/typesaurus, which will allow us to do that.

functions/src/models/users/index.ts
sabind: There’s an easier way to do this :)

functions/src/models/organizations/teams.ts
sabind: This is just a placeholder so that we can leave some mental room for orgs (and teams inside orgs) when thinking about users and other models that are used in the MVP.

functions/src/models/organizations/index.ts
sabind: `mututalAid=‘mutual_aid’`

idk either but thinking beyond Covid here in the US it’s called welfare so I imagine there is some distinction. We could probably leverage a social worker or someone to help us with the nomenclature?

functions/src/models/organizations/index.ts
sabind: In the code we won’t use the `string with spaces` so this is more for the front end team to be able to render? But they could also use a converter for the enum so.... 😅

functions/src/requests/index.ts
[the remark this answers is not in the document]
sabind: I think In understand better your comment now. there are two rating events on a request, but, maybe we average them into one number on the profile?

functions/src/offers/index.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  return Promise.all(operations);
+};
+
+const validateOffer = (value: IOffer): Promise<void> => {
+  return validate(Offer.factory(value))
+    .then(() => {
+      return Promise.resolve();
+    });
```
s0, line 33: Just realised that `validate` doesn't actually reject when it sees invalid data... you need to use `validateOrReject` for that...

```suggestion
  return validateOrReject(Offer.factory(value))
```

Though let's not worry about that right now as i'm working on pulling the model out into its own package anyway.

sabind: The unit tests pass relying on validate? It throws an error that is caught.

sabind:
@s0 If you approve feel free to merge so we unblock the front-end / module work :)

PR #244, opened 2020-04-01
Improve responsiveness of map

### LINKED ISSUE #75: Make Map UI Responsive
Right now, the [map](https://map.reach4help.org/) works quite well for tablet devices, and some phones in landscape, but is not designed to work well for mobile devices in portrait (which is probably going to be the majority of our users, if we play our cards right)!

Right now we just display a message to users with particularly narrow screens..

I was thinking the best way to go about improving this was to change the layout below the "mobile width threshold" so that the **header** and **results** views act as drawers that slide in from the top and bottom respectively, a bit like the google maps android app.

I think for now we can just deal with having buttons to open and close the drawers, but we'll probably want to evolve this to allow users to touch and drag these drawers open / closed, that can be a v2 though.

The best way to do this would probably be to create a new component for the drawer, that we wrap the results and header in. We'll probably want to either integrate the current open / close functionality of the results in this new component, or make it style-less above the mobile threshold.

I'm of course open to other ideas regarding the responsiveness here :)

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/75)

**Belonging to Epics:**

* **[EPIC]** [[@isma-osy](https://github.com/isma-osy)] [#413 - Prepare MAP for Go-To-Market / Marketing strategy](https://github.com/reach4help/reach4help/issues/413)

s0:
Description
-----------

Leaving this PR open while I work on the responsive design for the map, to get deployments on netlify to test on my phone :)

Don't merge yet! But ideas and feedback welcome in this PR.

TODO:

* [ ] vertical card layout for small-width devices
* [x] collapsible drawer for medium / large width devices

Motivation
----------

Testing Guidelines
------------------

Try it on mobile and tablets etc...

Additional Notes
----------------

Related to #75

### DIFF
--- a/map/src/App.tsx
+++ b/map/src/App.tsx
@@ -30,6 +30,12 @@ interface State {
   searchInput: HTMLInputElement | null;
   addInstructionsOpen: boolean;
   fullScreen: boolean;
+  /**
+   * * open: (default) the results are open
+   * * closed: the results are closed
+   * * open-auto: the results are open because a point is selected
+   */
+  resultsMode: 'open' | 'closed' | 'open-auto';
 }
 
 class App extends React.Component<Props, State> {
@@ -43,6 +49,7 @@ class App extends React.Component<Props, State> {
       searchInput: null,
       addInstructionsOpen: false,
       fullScreen: false,
+      resultsMode: 'open',
     };
   }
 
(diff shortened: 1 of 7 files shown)


s0:
Mobile view is still not done yet, and I think that'll be a task that I finish tomorrow, but I think this PR has enough enhancements now that we should get this in as-is, and i'll add the other functionality tomorrow :)

So feel free to review now.

PR #245, opened 2020-04-01
Dan/wrap auth

sabind:
Description
-----------

Tidy up Auth PR

Motivation
----------

Just tweaking a few things from @ashwinkjoseph

Testing Guidelines
------------------

Release Checklist
-----------------

Security Checklist
-------------------

Additional Notes
----------------

### DIFF
--- a/web-client/package.json
+++ b/web-client/package.json
@@ -43,7 +43,10 @@
   },
   "scripts": {
     "start": "SKIP_PREFLIGHT_CHECK=true react-app-rewired start",
-    "build": "NODE_ENV=development react-app-rewired build",
+    "build": "react-app-rewired build",
+    "build:dev": "NODE_ENV='DEVELOPMENT' react-app-rewired build",
+    "build:staging": "NODE_ENV='STAGING' react-app-rewired build",
+    "build:prod": "NODE_ENV='PRODUCTION' react-app-rewired build",
     "test": "react-app-rewired test",
     "eject": "react-app-rewired eject",
     "lint": "eslint . --ext .js,.ts,.tsx && prettier --check \"./src/**/*.{ts,tsx}\"",
--- a/web-client/src/firebase/index.ts
+++ b/web-client/src/firebase/index.ts
@@ -1,5 +1,4 @@
-import * as firebase from 'firebase/app';
-import 'firebase/auth';
+import firebase from 'firebase';
 
 const config = {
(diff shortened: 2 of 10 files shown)


web-client/package.json
```diff
@@ -43,7 +43,10 @@
   },
   "scripts": {
     "start": "SKIP_PREFLIGHT_CHECK=true react-app-rewired start",
-    "build": "NODE_ENV=development react-app-rewired build",
+    "build": "react-app-rewired build",
```
sabind, line 46: This is mostly the section I need help with in terms of using env vars in react.

`web-client/src/firebase/index.ts` contains the logic that should leverage the env var to load the proper config.

web-client/src/firebase/index.ts
```diff
@@ -9,8 +8,8 @@ const config = {
     projectId: 'reach4help-dev',
     storageBucket: 'reach4help-dev.appspot.com',
     messagingSenderId: '664078449280',
-    appId: '1:664078449280:web:fad2c3b30771213e5f4465',
-    measurementId: 'G-WFTRYNEQBB',
+    appId: '1:664078449280:web:115f33485dfc48e35f4465',
+    measurementId: 'G-DKH4PLZ62L',
```
jpmarques66, line 12: I think it isn't a good idea to have these keys and ids publicly in our repo 🤔

jpmarques66: Maybe this can go into .env files, what do ya think?

jpmarques66: I talked with Joseph and he'll handle this 😄 . Also, I don't if it is a good idea to maintain these values in the future. I think we should refresh the app id's and keys after they get removed from the code base. Maybe @mdeous can support me here.

sabind: https://firebase.google.com/docs/hosting/reserved-urls

sabind: > I talked with Joseph and he'll handle this 😄 . Also, I don't if it is a good idea to maintain these values in the future. I think we should refresh the app id's and keys after they get removed from the code base. Maybe @mdeous can support me here.

Will @ashwinkjoseph be taking over the work directly in this PR?

jpmarques66: I think he'll be creating other PR.

PR #243, opened 2020-04-01
Firebase Integration and Facebook Sign In with Firebase

ashwinkjoseph:
Description
-----------
This PR adds firebase as a dependency in the web-client project.
    The Configuration file for configuring firebase as per the environment has been created, this file will export an initialized firebase App instance, firebase auth instance and a firestore instance
    The Configuration file can choose firebase configuration depending on the NODE_ENV, however I have manually put it to development for the time being
    To Demonstrate how to use firebase with redux, I have implemented Facebook Sing in with Firebase Authentication and have changed the onclick callback of the singin in button to call the action creator for firebase auth

Motivation
----------
https://github.com/reach4help/reach4help/issues/231
https://github.com/reach4help/reach4help/issues/231
Testing Guidelines
------------------

Release Checklist
-----------------

- [ ] This code has unit tests
- [ ] I have a plan to verify that these changes are working as expected after deploy
- [ ] I have a plan for how to revert, rollback, or disable these changes if things are not working as expected

Security Checklist
-------------------

This PR creates, modifies, or deletes:

- [ YES] API routes: API routes, parameters, or user authorization
- [ YES] Authentication: Authentication mechanism
- [ YES] Credentials: Server side credentials, or secrets in configuration / source code
- [ NO] Cryptography: Encryption, hashing, certificates, signatures, random numbers, etc.
- [ YES] User data: personal information handling, logs, error messages, etc.

Additional Notes
----------------

PR #248, opened 2020-04-02
refactor(web-client): improved mobile support, secured keys in .env and persistent user sessions

ashwinkjoseph:
This is an improvement over the pull request #243, This Commit adds the following features:
1. Added support for signInWithRedirect in a manner that the internals can be replaced with other OAuth systems without changing the interface later on easily
2. Added the logic for determining if a user is on Mobile or on Desktop and choosing between signInWithPopUp and signInWithRedirect methods
3. Separated the configuration into environment variables and added the env keys to .env.Example
4. Added the logic for persisting users and logging in users if user has already logged in previously

Description
-----------

Motivation
----------

Testing Guidelines
------------------

Release Checklist
-----------------

Security Checklist
-------------------

Additional Notes
----------------


web-client/src/containers/LoginContainer/LoginContainer.tsx
```diff
@@ -14,14 +20,33 @@ const LoginContainer: React.FC<LoginRedirectProps> = ({
   const token = useSelector((state: AppState) => state.auth.token);
   const history = useHistory();
 
+  const redirectStarted = window.localStorage.getItem('redirect_started');
```
jpmarques66, line 23: Do you think this can be a selector? Instead of saving this in LocalStorage directly, saving it in a reducer, what do you think?

web-client/src/ducks/auth/reducer.ts
```diff
@@ -20,11 +28,93 @@ export default createReducer<AuthState>(
...
+    },
+    [FIREBASE_FACEBOOK_LOGIN_REDIRECT_START.COMPLETED]: (state: AuthState) => {
+      const actionInProgress = false;
+      return {
+        ...state,
+        actionInProgress,
+      };
+    },
+    [FIREBASE_FACEBOOK_LOGIN_REDIRECT_COMPLETE.PENDING]: (state: AuthState) => {
+      window.localStorage.removeItem('redirect_started');
```
jpmarques66, line 88: Here we could just save `redirect_started` in reducer state perhaps?

web-client/src/http/resources/auth.ts
```diff
@@ -11,15 +11,28 @@ export const login = (request: IHTTPRequest) => {
...
 };
 
-export const facebookLoginWithFirebase = async (): Promise<firebase.auth.UserCredential> => {
+export const facebookLoginWithFirebasePopUp = async (): Promise<
+  string | undefined
+> => {
   const provider = new firebase.auth.FacebookAuthProvider();
   const result = await firebaseAuth.signInWithPopup(provider);
-  return result;
+  const token = await result.user?.getIdToken();
```
jpmarques66, line 19: In these request we could just continue using the promise, since `injectRequestMiddleware` already handles the `await` for us. I think this way is a bit cleaner. What do you think?

jpmarques66: Basically, just like we would do with axios.

web-client/src/ducks/auth/reducer.ts
```diff
@@ -20,11 +28,93 @@ export default createReducer<AuthState>(
     ) => {
       state.token = payload.accessToken;
     },
-    [FIREBASE_FACEBOOK_LOGIN.COMPLETED]: (
+    [FIREBASE_FACEBOOK_LOGIN_POPUP.PENDING]: (state: AuthState) => ({
+      ...state,
```
jpmarques66, line 32: Here we don't need to use spread operator. The state object is a draft, since we're using immerJs 🚀 . You can just go ahead ad mutate the state props.

web-client/src/containers/LoginContainer/LoginContainer.tsx
ashwinkjoseph: @jpmarques66 
no, the reason for this is that it's coming after a redirect. so you want it to persist.
You could have redux-persist or redux-offline. However, I don't think redux-persist or redux-offline is to be used for solving these issues. They should be integrated but for improving offline accessibility and improving UX, but that is something that I feel should be done at a later point in the development
This is, however, my opinion. If you think that this is wrong and we need to use redux-persis. then I will add that as another PR for redux-persist

web-client/src/ducks/auth/reducer.ts
ashwinkjoseph: my intention for redirect_started as a parameter in "localStorage" was for it to persist across browser sessions.
We can use this in the manner you are suggesting only if we are using redux-persist, which in my opinion shouldn't be used for the purpose of solving this particular issue only. and should be for improving user experience offline.
However as I have mentioned in the previous comment, this my opinion.
I am open to your point of view and am willing to add redux-persist but that would be done as a separate PR

web-client/src/http/resources/auth.ts
ashwinkjoseph: oh yes, that makes sense.. now I understand why the error actions weren't being dispatched. 
However, only the promise for the token should be handled by it. I will make the change in PR.
I did go through the middleware, but maybe because I was trying to solve another issue. this particular point of view skipped my mind.
I will make the change

web-client/src/ducks/auth/reducer.ts
ashwinkjoseph: okay, I wasn't aware of that. I will make the change

web-client/src/containers/LoginContainer/LoginContainer.tsx
sabind: > Private Browsing / Incognito modes

> Most modern browsers support a privacy option called 'Incognito', 'Private Browsing' or something similar that doesn't store data like history and cookies. This is fundamentally incompatible with Web Storage for obvious reasons. As such, browser vendors are experimenting with different scenarios for how to deal with this incompatibility.

> Most browsers have opted for a strategy where storage APIs are still available and seemingly fully functional, with the one big difference that all stored data is wiped after the browser is closed. For these browsers there are still different interpretations of what should be done with existing stored data (from a regular browsing session). Should it be available to read when in Private mode? Then there are some browsers, most notably Safari, that have opted for a solution where storage is available, but is empty and has a quota of 0 bytes assigned, effectively making it impossible to write data to it.

> Developers should be aware of these different implementations and take them into account when developing websites depending on Web Storage APIs. For more information please have a look at this WHATWG blog post that specifically deals with this topic.

Just a thought — but I do agree that for now any persistence is good, but I’d recommend sessionStorage since we don’t need longevity or a large amount of storage.

sabind: Or we can use a cookie, whichever works in the most cases again since we’re storing almost no data and only need it for a short period of time.

web-client/src/ducks/auth/reducer.ts
ashwinkjoseph: I have made this change now

web-client/src/http/resources/auth.ts
ashwinkjoseph: I have made this change now

web-client/src/containers/LoginContainer/LoginContainer.tsx
ashwinkjoseph: @sabind @jpmarques66 
This is not actually required. I could directly call firebaseAuth.getRedirectResult when the component is loaded, but the problem is, it takes a while for the callback to be called.

The problem with this is, suppose you login with redirect and are redirected back to the website, there will be a lag between the site loading and the firebase SDK calling the callback which let's us do further processing and take the user to the protected route.

The user would be clueless as to what is happening as the user had already logged into the system and our web app wouldn't be knowing that it actually came back from an OAuth Consent screen. 

My intention was to have a loading screen in between this and for that I used a persistent storage to let the app know that something is happening.

This is not vital to the functionality of the app and I can remove it if needed

ashwinkjoseph: Due to this discussion, I ended up reading more about these. it turns out a lot of things have to be taken into consideration regarding the nature of the data that we are storing when we choose between cookies and a Storage API.

https://stackoverflow.com/a/36367799/5356465

But none of this affects the problem this particular code snippet addresses, even if it's not set on private browsing, it just means that the user has to have that bad experience of waiting without knowing what's happening.

If we were to move away from firebase at some later point of time and were to accept and deal with the access token ourselves as an OAuth Client, we wouldn't have this problem as the web app would be knowing that it's a redirect from the consent screen, just by identifying the query parameters

micaelr95:
@all-contributors please add @jpmarques66 for review.

PR #270, opened 2020-04-05
Adding timeline item

sabind:
Description
-----------

The timeline object can be used to track the request as well as the active offer (in case an offer is accepted or rejected along with the status change).

Motivation
----------

@puzzledbytheweb pointed out that UI can't render a timeline view if we don't track status changes of the request with a timestamp.

Testing Guidelines
------------------

Release Checklist
-----------------

Security Checklist
-------------------

Additional Notes
----------------

### DIFF
--- a/functions/src/models/requests/timeline.ts
+++ b/functions/src/models/requests/timeline.ts
@@ -1,0 +1,96 @@
+import { FirestoreDataConverter } from '@google-cloud/firestore';
+import { IsEnum, IsObject } from 'class-validator';
+import { firestore } from 'firebase-admin';
+
+import { IOffer } from '../offers';
+import { IRequest, RequestStatus } from './index';
+import Timestamp = firestore.Timestamp;
+import DocumentData = firestore.DocumentData;
+
+export interface ITimelineItem extends DocumentData {
+  offerSnapshot: IOffer;
+  requestSnapshot: IRequest;
+  status: RequestStatus;
+  createdAt?: Timestamp;
+}
+
+/**
+ * Cannot be created/modified/deleted by a User. This is audit log maintained by the system
+ * It is used to render a timeline of the request on the front end, but all records
+ * are created automatically by the system based on status changes.
+ */
+export class TimelineItem implements ITimelineItem, FirestoreDataConverter<TimelineItem> {
+
+  @IsObject()
+  private _offerSnapshot: IOffer;
+
+  @IsObject()
+  private _requestSnapshot: IRequest;
+
+  @IsEnum(RequestStatus)
+  private _status: RequestStatus;
+
+  @IsObject()
+  private _createdAt: Timestamp;
+
+  constructor(offerSnapshot: IOffer, requestSnapshot: IRequest, status: RequestStatus, createdAt: Timestamp) {
+    this._offerSnapshot = offerSnapshot;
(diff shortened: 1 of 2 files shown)


sabind:
I should add that the front end should query both `request.offers` and `request.timeline` to get all the information it needs (it can zip the data together based on timestamp)

puzzledbytheweb:
Looks great, but I may be missing something or maybe I just need guidance on how to make the query. So the file that was created shows the TimelineItem Type, along with the functions to interact with Firebase. As such, I'm just wondering if the Request Type/Model has the references for this TimelineItems and if this is organized neatly as an array of changes (that we can order by timestamp). If it's not, or it is not possible due to the nature of the Firebase database, should I do `find` the TimelineItems by RequestId? I do have that data on that page but I'm afraid I don't have the OfferId in the PIN part of the app.

functions/src/models/requests/timeline.ts
```diff
@@ -0,0 +1,96 @@
...
+
+import { IOffer } from '../offers';
+import { IRequest, RequestStatus } from './index';
+import Timestamp = firestore.Timestamp;
+import DocumentData = firestore.DocumentData;
+
+export interface ITimelineItem extends DocumentData {
+  offerSnapshot: IOffer;
+  requestSnapshot: IRequest;
+  status: RequestStatus;
```
s0, line 13: `status` is not necessary if it's included in `requestSnapshot` right? which will include even more information?

Unless `requestSnapshot` is the (previous) snapshot? But i think it makes more sense to have it be the *new* state of the request...

We would need to make sure that we create a timeline item for the initial creation of requests in that case (should be pretty easy with an event)

```suggestion
```

functions/src/models/requests/timeline.ts
```diff
@@ -0,0 +1,96 @@
...
+import { firestore } from 'firebase-admin';
+
+import { IOffer } from '../offers';
+import { IRequest, RequestStatus } from './index';
+import Timestamp = firestore.Timestamp;
+import DocumentData = firestore.DocumentData;
+
+export interface ITimelineItem extends DocumentData {
+  offerSnapshot: IOffer;
+  requestSnapshot: IRequest;
```
s0, line 12: I'm assuming that from the file path, you're planning on having these routed as sub-collections under requests, which allows us to easily query the timeline for a single request, but not across requests.

This makes sense to me .

sabind: Yes, it's a sub of each request so you can ask a request for it's timeline and all it's offers :)

functions/src/models/requests/timeline.ts
```diff
@@ -0,0 +1,96 @@
...
+  private _status: RequestStatus;
+
+  @IsObject()
+  private _createdAt: Timestamp;
+
+  constructor(offerSnapshot: IOffer, requestSnapshot: IRequest, status: RequestStatus, createdAt: Timestamp) {
+    this._offerSnapshot = offerSnapshot;
+    this._requestSnapshot = requestSnapshot;
+    this._status = status;
+    this._createdAt = createdAt;
```
anandvenkat4, line 35: Do we have to assign now() for createdAt in the constructor itself, if the value is not supplied?

sabind: Good point! Yes we should do this for the user.

Issue #100, opened 2020-03-23
Implement view: Name & Address


comoser:
- pre-polutate the most info possible from the info requested to `facebook`
- request user address if not able to retrieve from facebook
- address should be composed of `Address`, `City`, `Country`

- [x] Build dummy visual page component
- [x] Update visuals for mobile
- [x] Integrate with Redux & Firebase

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/100)

*This information is updated automatically. To modify it, please use ZenHub.*

**Belonging to Epics:**

* ~~**[EPIC]** [#327 - View: Name & Address](https://github.com/reach4help/reach4help/issues/327)~~

**Blocked By:**

* ~~[#100 - Design view: Name & address](https://github.com/reach4help/reach4help/issues/344)~~

PR #285, opened 2020-04-07
Feat/#100 user data form

micaelr95:
Description
-----------

Personal data form.

Motivation
----------

Testing Guidelines
------------------

Release Checklist
-----------------

Security Checklist
-------------------

Additional Notes
----------------

Issue #100 

image


s0:
I won't have time to review the code, but I do have one comment regarding the UX (which is probably more for @telmodias or @cinthyaejh). "First name" and "Last name" is a little bit too restrictive, and doesn't apply for many people / a number of different cultures. Some people only have one name, some people have more than 2 and it doesn't make sense to split them etc...

It's generally better to just ask for "full name" with a single text field, so can we do that?

schowdhury8:
Agree with Sam about the name thing (then to personalize either do a split()[0] and address them by their first name or give them a nickname option?), the design looks amazing and really clean, but I do wanna reiterate the point of designing this with mobile responsiveness in mind (haven't had a chance to test it myself). Also maybe the font could be a bit bigger. That's all from me, fantastic job everyone!!

telmodias:
I was reading about this matter with the first/last names vs full name. My point with full name, is that i.e. people in Portugal have many names, my full name is Telmo de Campos Ferreira Dias .
My main point is in order to ease communication with the user, often you might use things as "Welcome back Telmo!" or on an e-mail "Hello Telmo, (...)". No imagine a user experience on a mobile device saying: Hello, Telmo de Campos Ferreira Dias!

I was reading https://blog.prototypr.io/full-name-vs-first-last-name-ux-best-practice-8c2db7178fd1 and https://ux.stackexchange.com/questions/20513/is-it-better-to-ask-for-full-name-or-first-name-and-last-name-in-a-registr . My favorite is 1st and last because it would make it the most simple to implement and to have in a form (last name could be optional). The other cases I would go with having another field which could be "Display Name" (3rd reply on the ux.stackexchange.com link).

The minimal option for most inclusiveness could be : full name, display name, but then as the other article says, if you ask me these 2, you should explain to me (GDPR ?) why you need my full name.

Other option is to go only with display name, but for official records like for the future usage within organizations, I think full name might be important, but maybe we implement this further down the road ?

Pease tell me what you think.

BTW, German way of dealing with this is crazy :

1st name: Telmo
last name: de Campos Ferreira Dias .

This is what the bank ppl did, and then the last name doesn't have enough string length so my name never shows the s of Dias.

s0:
@telmodias good points.

In that case, i think it's best to move forward to begin with with just "Display Name" or "Name". If we require full names for organizations, we can add that later on?

> BTW, German way of dealing with this is crazy :
> 
> 1st name: Telmo
> last name: de Campos Ferreira Dias .
> 
> This is what the bank ppl did, and then the last name doesn't have enough string length so my name never shows the s of Dias.

Oh god :grimacing:

telmodias:
My preferred would be 1st and last, being last optional, just because : 

1. you can safely say "Hi Sam"
2. you can safely write a letter to Sam Lanning 

My point is if we leave it only name or display name, then we can't mail things safely to their address.

What are your views on this @schowdhury8 @micaelr95 @jpmarques66 @ashwinkjoseph  ?

schowdhury8:
@telmodias @s0 I think just to be safe, it's a good idea to have full name and display name which defaults to the first word of their full name. Collecting the full name would be useful for reporting to the organizations they're a part of and potentially if one of the volunteers needs to identified for whatever reason (foul play, etc.) It is an extra thing to add to our GDPR Privacy Policy but I have a feeling it'll be very important from a legal liability standpoint.

s0:
> @telmodias @s0 I think just to be safe, it's a good idea to have full name and display name which defaults to the first word of their full name. If need be, we could make the display name be the only thing publicized. Collecting the full name would be useful for reporting to the organizations they're a part of and potentially if one of the volunteers needs to identified for whatever reason (foul play, etc.) It is an extra thing to add to our GDPR Privacy Policy but I have a feeling it'll be very important from a legal liability standpoint.

I like this, i agree! :+1:

telmodias:
Ok so be it. @micaelr95 please take note :
1st field for Full Name
2nd field for Display Name (value defaults to 1st word in Full Name).

web-client/src/pages/routes/UserDataFormPage.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+import CenteredCard from '../../components/CenteredCard/CenteredCard';
+import GradientBackground from '../../components/GradientBackground/GradientBackground';
+import UserDataForm from '../../components/UserDataForm/UserDataForm';
+
+const UserDataFormPage: React.FC = () => {
```
jpmarques66, line 7: This should be named route instead of page

jpmarques66: Refer to https://github.com/reach4help/reach4help/wiki/Web-client#routes

web-client/src/components/UserDataForm/UserDataForm.tsx
```diff
@@ -1,5 +1,6 @@
 import { Button, Checkbox, Form, Input, Typography } from 'antd';
-import React from 'react';
+import _ from 'lodash';
```
jpmarques66, line 2: Here you can import `words` from lodash like so:

`import words from 'lodash/words';`

This will reduce bundle size as we not importing lodash's full package.

micaelr95: Can I import like this? `import { words } from 'lodash';`

micaelr95:
Changed inputs to full name and display name. The display name is filled automatically with the first word of the first name.

image

PR #292, opened 2020-04-07
feat(map): change default options in map creation

### LINKED ISSUE #287: Center map on startup!
Right now, the default coordinates (0, 0) are not very useful for many resolutions, and most of the markers start off the end of the map.

We should start the map at the center of the markers we know about instead, something like:

* lat: `(max(lat) + min(lat)) / 2)`
* lng: `(max(lng) + min(lng)) / 2)`

Relevant lines here: https://github.com/reach4help/reach4help/blob/master/map/src/components/map-utils/google-maps.ts#L7

btuerker:
Description
-----------
Changed the default coordinates in the options parameter. Used the average of both latitude and longitude values.

Motivation
----------

Testing Guidelines
------------------

Release Checklist
-----------------

Security Checklist
-------------------

Additional Notes
----------------
Visual changes are already presented in issue #287 

closes #287

### DIFF
--- a/map/src/components/map-utils/google-maps.ts
+++ b/map/src/components/map-utils/google-maps.ts
@@ -1,10 +1,19 @@
+import { MARKERS } from '../../data/markers';
 import { parseQueryString } from './query-string';
+
+const average = (arr: Array<number>) =>
+  arr.reduce((acc, curr) => acc + curr) / arr.length;
 
 export const createGoogleMap = (ref: HTMLDivElement): google.maps.Map => {
   const query = parseQueryString();
   return new google.maps.Map(ref, {
     zoom: query.map ? query.map.zoom : 3,
-    center: query.map ? query.map.pos : { lat: 0, lng: 0 },
+    center: query.map
+      ? query.map.pos
+      : {
+          lat: average(MARKERS.map(m => m.loc.lat)),
+          lng: average(MARKERS.map(m => m.loc.lng)),
+        },
     mapTypeId: google.maps.MapTypeId.ROADMAP,
     streetViewControl: false,
     clickableIcons: false,


s0:
Perfect, thanks!

btuerker:
Also, I'd like to mention some problems that may occur.

1. Currently, the average function I've implemented is not secure to use it with an empty array. So if we send an empty array into it, it's giving error. I've just realized after the push.

```javascript
average([]) // -> throws TypeError: Reduce of empty array with no initial value
```
2. At the MARKERS.map() process, I really don't know it's already type-safe right now. I've pretty limited typescript knowledge. Can we ensure that each type specified in its interface will be there? I mean, is there any need to implement something like:
```javascript
MARKERS.map(marker => marker.loc ? marker.loc.lat : 0)
```

s0:
> Also, I'd like to mention some problems that may occur.
> 
> 1. Currently, the average function I've implemented is not secure to use it with an empty array. So if we send an empty array into it, it's giving error. I've just realized after the push.
> 
> ```js
> average([]) // -> throws TypeError: Reduce of empty array with no initial value
> ```

For the average not working with an empty array, i think that's okay for now as it is hard-coded and non empty, in any case you can't average an empty array, that'll be NaN, so throwing an error in this case actually makes sense to me.




> 1. At the MARKERS.map() process, I really don't know it's already type-safe right now. I've pretty limited typescript knowledge. Can we ensure that each type specified in its interface will be there? I mean, is there any need to implement something like:
> 
> ```js
> MARKERS.map(marker => marker.loc ? marker.loc.lat : 0)
> ```

`marker.loc` is guaranteed to be defined, if it wasn't it would have produced a type error, so the additional check of the nullness of `loc` is not necessary :)

micaelr95:
@all-contributors please add @btuerker for code.

PR #293, opened 2020-04-07
Update issue templates

sabind:
These are simply the default issue templates, but I think we should start thinking about what might work best.


s0:
These look good to begin with :)

micaelr95:
@all-contributors please add @sabind for doc.

Issue #288, opened 2020-04-07
Center map based on users country on startup


s0:
As a follow-up to #287,

it would be great if we could determine the **rough** country (or other locale) of the current user, and begin the maps centered there. This shouldn't require the qeolocation API, as requiring a user's geo on app startup is quite invasive, instead doing it based on IP address or something would be sufficient.

We'd need some backend service for this, either our own cloud functions on firebase (or netlify, but we're using firebase for all other backend things), or some third party like MaxMind (which is expensive, but there may be other options).

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/288)

*This information is updated automatically. To modify it, please use ZenHub.*

PR #343, opened 2020-04-08
feat(map): add center to my location button

### LINKED ISSUE #289: Allow users to center on their location
We should have a button that centers the map on the user's current location, by requesting their location using the javascript geolocation api.

This functionality is probably built in to the maps javascript sdk, but I haven't looked into it.

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/289)

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/289)

btuerker:
## Description
Added another button to the bottom of map. Created a wrapper for styling the group of buttons. Implemented handler function.
## Release Checklist

## Security Checklist

## Additional Notes
Here're the visual changes:
Screen Shot 2020-04-08 at 5 54 16 PM
Screen Shot 2020-04-08 at 5 54 27 PM

closes #289

### DIFF
--- a/map/src/components/map.tsx
+++ b/map/src/components/map.tsx
@@ -1,6 +1,11 @@
 import isEqual from 'lodash/isEqual';
 import React from 'react';
-import { MdExpandLess, MdExpandMore, MdRefresh } from 'react-icons/md';
+import {
+  MdExpandLess,
+  MdExpandMore,
+  MdMyLocation,
+  MdRefresh,
+} from 'react-icons/md';
 import { Filter, SERVICES } from 'src/data';
 import { button, iconButton } from 'src/styling/mixins';
 
@@ -366,6 +371,19 @@ class MapComponent extends React.Component<Props, {}> {
     }
   };
 
+  private centerToGeolocation = () => {
+    navigator.geolocation.getCurrentPosition(position => {
+      const pos = {
+        lat: position.coords.latitude,
+        lng: position.coords.longitude,
+      };
+      if (!this.map) {
+        return;
+      }
+      this.map.map.setCenter(pos);
+    });
+  };
+
   private initializeSearch() {
     const { searchInput } = this.props;
     if (this.searchBox?.searchInput !== searchInput) {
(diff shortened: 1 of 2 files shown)


sabind:
When we center on location we should run the "Update results for this area" function as well.

map/src/components/map.tsx
```diff
@@ -366,6 +371,19 @@ class MapComponent extends React.Component<Props, {}> {
...
+  private centerToGeolocation = () => {
+    navigator.geolocation.getCurrentPosition(position => {
+      const pos = {
+        lat: position.coords.latitude,
+        lng: position.coords.longitude,
+      };
+      if (!this.map) {
+        return;
+      }
+      this.map.map.setCenter(pos);
```
sabind, line 383: When we center on location we should run the "Update results for this area" function as well.

sabind:
👍 Look really good! Made one comment, and then I think it's ready to go. @s0 may have some code quality review comments (I'm not super familiar with react)

map/src/components/map.tsx
s0: Agreed, we should perhaps also modify the zoom level

btuerker: I got stuck at updating results. Simply `this.setResults()` doesn't work. Because, after I set the map's center it doesn't change the `nextResults`  immediately. It needs to wait for cluster calculation, then updates the results. I couldn't find a proper way to do it. 

I've just tried to set a timeout for some milliseconds then only after calling `updateResults`, it works but it really is a poor solution.

I'm looking forward to hear your ideas because this problem will cause the same issue when we're implementing the follow-up feature #288 

The event `MarkerClusterer - 'clusteringend'` is responsible for setting nextResults after the bounds are changed. As shown below.
https://github.com/reach4help/reach4help/blob/78b6ee21901f9de72848c692e53a3ae58f66e083/map/src/components/map.tsx#L292-L300

s0: for this, i would introduce a boolean property called `updateResultsOnNextClusteringChange` that you set to true before changing the map coordinates, then check that value in the `clusteringend` event listener, and if true, set it to false and call `this.setResults` :)

btuerker: Eureka!!!

btuerker:
The latest commit also changes the map zoom. I set it to 8. Here's visual: 
Screen Shot 2020-04-09 at 7 27 47 PM

map/src/components/map.tsx
```diff
@@ -366,6 +385,22 @@ class MapComponent extends React.Component<Props, {}> {
     }
   };
 
+  private centerToGeolocation = () => {
+    navigator.geolocation.getCurrentPosition(position => {
```
s0, line 389: one final comment, let's also add an error handler that prints the error to the console, and just uses an `alert()` to display a message like "Unable to get geolocation".

btuerker: I'm using chrome too and it's working :) Let's see the error prompt then we can think through. It's probably a permission error. If then; we'll need to implement something like: 
```javascript
navigator.permissions.query({ name: 'geolocation' })
  .then((result) => {
    if (result.state == 'granted') {
     ...
    } else if (result.state === 'prompt') {
      ...
    } else if (result.state === 'denied') {
      // user must enable it manually once denied
      alert('Ask for to give permission manually')
    }
 });
```

s0: I'm actually getting an unusual response:

```
Network location provider at 'https://www.googleapis.com/' : No response received.
```

In any case, working well now i think!

PR #376, opened 2020-04-10
Improvements to ZenHub issue sync

s0:
## Description

* Include which epics an issue is part of
* Remove some spurious curly braces

### DIFF
--- a/.github/zenhub-issue-sync/dist/index.js
+++ b/.github/zenhub-issue-sync/dist/index.js
@@ -3228,10 +3228,10 @@ var REPO_EXTRACT = /^([^\/]+)\/([^\/]+)$/;
 var START_BOUNDARY_COMMENT = "\n\n<!--zenhub info:";
 var END_BOUNDARY_COMMENT = "<!--zenhub info end-->";
 (function () { return __awaiter(void 0, void 0, void 0, function () {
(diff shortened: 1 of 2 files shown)


ashwinkjoseph:
> I have a question regarding these variable names (e, e_1, e_2, etc). Do you think that this can hinder future collab on this particular service or is it okay to leave it like this? I find it a little difficult to understand what it represents. But other than that it looks good to me!

Hey, are you checking the compiled .js or are you checking the .ts file?

s0:
Yeah that's just the compiled JS, it needs to be checked in because it's running as a github action :)

s0:
> I have a question regarding these variable names (e, e_1, e_2, etc). Do you think that this can hinder future collab on this particular service or is it okay to leave it like this? I find it a little difficult to understand what it represents. But other than that it looks good to me!

puzzledbytheweb:
> > I have a question regarding these variable names (e, e_1, e_2, etc). Do you think that this can hinder future collab on this particular service or is it okay to leave it like this? I find it a little difficult to understand what it represents. But other than that it looks good to me!
> 
> Yeah that's just the compiled JS, it needs to be checked in because it's running as a github action :)

Ah thanks, was not understanding that ;)

PR #377, opened 2020-04-10
Map Dataset

Tahfimul:
## Description
Created JSON dataset for Map
## Motivation
Map required data collection from various sources, previously hard coded data under markers.ts file needed to be replaced by a JSON file, a more accessible data file format. Future marker data can be aggregated to this JSON file which can eventually merge into Firebase and data could then be fetched from Firebase to Map.
## Testing Guidelines
In order to make use of this JSON file, map.ts code has to be restructured to parse each marker data from the JSON file.
## Release Checklist
head, 17 comm

## Security Checklist


sabind:
@s0 I like this change where people can add individual files so that we don't have conflicts on the single one large file 👍

Tahfimul:
> Made a bunch of changes that load the json file using fetch (instead of using the import, which wasn't working for some reason), and allowing async loading of the markers, which we can take advantage of when we load that data using firebase.

I really like this forward thinking plan. I think this is a better implementation.

Tahfimul:
> I also reverted the change pulling the `MARKERS` constant into a class. I can see the logic of why you did it (as the value was modified by the class). Unfortunately that was causing a conflict with other changes that have occurred, so it made sense to leave it where it was, and not modify it.

That's great!

Tahfimul:
> * There are lots of descriptions that have the text `&endl;` in them, i'm guessing this is either supposed to be a line break or paragraph end? Could we make the appropriate replacement.

Sure thing! That will be in the works.

Tahfimul:
> * All the values seem to have the exact same services (all of them), and don't in any way seem to relate to what's actually provided. Can we do some sort of cleanup here? Maybe i could help suggest something if you share what the source of this information was again?

About this last point, I pulled the dataset from this link: https://mutualaid.nyc/groups/ . Since many of the organizations did not provide keywords to their services which fall into one of our `Services` category, I chose to put them all under all of our `Services` tags. Moving forward, I will be putting more emphasis on this factor as well. Thanks for pointing it out!

sabind:
> Although, I feel that if we include all datasets for the markers into a unified file, loading the data would be more efficient. But I side with whichever you feel is better.

If we load multiple at once we can complete that operation in parallel for some of the operations. Since we're serving via CDN I'm not too concerned with having to conduct too many handshakes for small datasets (vs one for a huge dataset).

Future changes will store the data into a DB so this optimization should be made for ease of use vs. performance.

sabind:
> Let me know how I can best structure it or if I should include the proper `<br>` syntax for HTML instead for these kinds of datasets.

These are HTML blocks so we should stick to HTML syntax.

s0:
I think we need to put a hold on getting this in until #414 is resolved, and then rework the data a little bit more to get the info in the correct format. I'll be using inspiration from the NYC mutual aid page to inform the data model, so it'll probably much more closely resemble the original data too, and we won't have to abuse "Services" quite so much...

To be honest though, I am somewhat hesitant about including the dataset from mutualaid.nyc in here ([see my thoughts here](https://github.com/reach4help/reach4help/wiki/Map-Project:-Unifying-Mutual-Aid-Group-Directories-and-Maps#map-nyc-mutual-aid-groups)), and instead think that we should direct people to mutualaid.nyc when they're looking at that area (see #415).

I'm open to being persuaded that we include a snapshot of this data AND also direct people to that map too (given it's functionality is currently much better and more granular than ours).

s0:
Hi @Tahfimul, i just went ahead and cleaned up some files in your PR that I don't believe you intended to change.

In addition, i removed your changes to the README for now as i'm just going to replace that with instructions to use the map UI to submit data.

I am keen to get some of this data in though, but it's still out of date with the current schema (the "type" fields need to be updated), there are also 2 json files currently, what's the difference between them? and where does this data come from? It looks like it's information on food banks for children mostly, but it's not clear.

I think before we get this merged in, we should also add a "source" field to these entries so we don't lose track of where they came from :)

schowdhury8:
Closing because stale. 
@Tahfimul @s0 if you still have interest in including this data, feel free to update and reopen it!

PR #380, opened 2020-04-10
[RFC][#378] - Split reducer logic

### LINKED ISSUE #378: Separate profile and auth reducers
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/378)

jpmarques66:
## Description
This PR splits reducer login related to auth, to allow for better code management.
## Release Checklist

## Security Checklist

## Additional Notes

Closes #378


web-client/src/ducks/auth/facebook/types.ts
```diff
@@ -0,0 +1,14 @@
+import createActionTypeFactory from 'src/store/utils/createActionTypeFactory';
+
+const { asyncType: firebaseAsyncType } = createActionTypeFactory('FIREBASE');
```
ashwinkjoseph, line 3: I was wondering, should this "type factory" be defined as Firebase or as Auth?
Since we are any way implementing a hierarchy, shouldn't this be the name of the module? or at least <Module Name>/<Sub Module or Reducer Name>/type

web-client/src/ducks/auth/facebook/actions.ts
```diff
@@ -0,0 +1,33 @@
+import {
+  facebookLoginWithFirebasePopUp,
+  getRedirectResult,
+  loginWithFirebaseRedirect,
+} from 'src/http/resources/auth';
```
ashwinkjoseph, line 5: I was wondering, these functions are the operations that are to be carried out for an `action creator` and since these functions are again created as modules (because all the functions related to auth are put in auth.ts in resources directory they wouldn't really be used again for an `action creator`), wouldn't it be better for us to define these in the same directory as the action creators in a file called functions.ts? 
I also think that putting all auth related functions in `src/http/resources/auth` is valid only for the old manner of redux modules now we have sub redux modules, so I think these should also be having hierarchy)

web-client/src/modules/login/containers/LoginContainer/LoginContainer.tsx
```diff
@@ -2,11 +2,12 @@ import React, { useEffect } from 'react';
 import { isMobile } from 'react-device-detect';
 import { useDispatch, useSelector } from 'react-redux';
 import { useHistory } from 'react-router-dom';
+import {} from 'src/ducks/auth/actions';
```
ashwinkjoseph, line 5: Shouldn't this be removed?

web-client/src/modules/phone/pages/ContentPage.tsx
```diff
@@ -12,7 +12,7 @@ import { PhoneVerifyLocation } from './routes/PhoneVerifyRoute/constants';
 import PhoneVerifyRoute from './routes/PhoneVerifyRoute/PhoneVerifyRoute';
 
 const ContentPage = (): ReactElement => {
-  const user: firebase.User = useSelector((state: AppState) => state.auth.user);
+  const user = useSelector((state: AppState) => state.auth.user);
```
ashwinkjoseph, line 15: shouldn't we enforce type here? it would help us while accessing the properties right?

jpmarques66: No, I refactored the store implementation and, right now, you don't need to cast the type of state properties. They are already recognized of being of that type.

web-client/src/modules/login/containers/LoginContainer/LoginContainer.tsx
jpmarques66: yes 😅 thanks

jpmarques66:
> I don't think merging the sub reducers from Facebook and phone as done in ([src/ducks/auth/reducer.ts](https://github.com/reach4help/reach4help/blob/feat/%23378-separate-reducers/web-client/src/ducks/auth/reducer.ts)) is the right way to approach the issue we have at hand right now.
> 
> This method only addresses the issue of how to split the code for the phone, Facebook and so on into separate files while being in the same Redux Module.
> 
> The problem that we were trying to address was how to manage multiple reducers for the same module.
> An example of this would be when we would need a reducer to store details about a request to create a document and another to store details that we get from creating a document.
> 
> We were thinking of using case functions in reducers right? However, looking at this PR, I think we should try another approach.
> 
> The nature of the problem we discussed above is such that the actions performed or recorded by the reducers are different in nature and shouldn't interfere with one another.
> However, as seen in this PR, there is also a problem where we would need to deal with multiple different operations that would need to deal with the same state as here all forms of authentication still needs to use the same state variables (user and token) and so must be allowed to change these states.
> 
> So we would need to see if the `case function` method described in redux docs is the right way to approach this and how we can implement this using it.
> 
> However, we could make this change by assigning the value of each submodule to a property in the state, thus dividing the state into multiple attributes and having each submodule be responsible for one attribute. This way the submodules would be able to access the same state as the parent module and would also be able to implement isolation or the 1st case I mentioned.

Has discussed with @ashwinkjoseph through PM, we will follow this approach for now and when we get to a point where we can better split the ruder properties we'll split this better.

web-client/src/ducks/auth/facebook/types.ts
jpmarques66: That makes sense!

web-client/src/ducks/auth/phone/types.ts
```diff
@@ -0,0 +1,18 @@
...
+export const FIREBASE_PHONE_VERIFY = asyncType('PHONE_VERIFY');
+
+export interface IPhoneNumberAuth {
+  phoneNumber: string;
+  currentUser?: firebase.User;
+  // eslint-disable-next-line @typescript-eslint/camelcase
+  recaptchaVerifier: firebase.auth.RecaptchaVerifier_Instance;
+}
+
+export interface IOTPAuth {
```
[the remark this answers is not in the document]
jpmarques66: That makes sense. I didn't notice I was using them outside /phone

web-client/src/ducks/auth/facebook/actions.ts
jpmarques66: I split some of the http functions into different files. Although we should think of a better way fo splitting this.

web-client/src/modules/phone/pages/ContentPage.tsx
ashwinkjoseph: okay sounds good

web-client/src/ducks/auth/phone/types.ts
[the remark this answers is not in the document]
ashwinkjoseph: what I meant was that the type should be defined in the same file as the file in which the action creator is defined.
you have defined all action creator types in the main types of the redux module. 
This still has the same issue I mentioned above, but now also has the issue of too much code in one place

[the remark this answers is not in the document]
jpmarques66: So, you mean doing this in actions.ts?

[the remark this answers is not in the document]
jpmarques66: I've been thinking about these types, I think we should try to abstract them into a directory in our project. 
Except for Interfaces that are very specific, such as an interface to represent a component's props, we should declare these interfaces in a more common space because we never know where we're going to need them.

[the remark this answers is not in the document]
ashwinkjoseph: yes I mean in actions.ts
but the thing is interfaces are mostly going to be needed when we are interacting with an action creator.
and the interface for interacting with the action creator may not be the interface when interacting with the actual function as seen in the verify OTP action creator.
so I think defining them along with the actions creators in the actions.ts would be the best solution

web-client/src/ducks/auth/phone/actions.ts
```diff
@@ -0,0 +1,29 @@
+import {
+  phoneAuthTrigger,
+  phoneAuthVerify,
+} from 'src/http/resources/auth/phoneAuth';
```
ashwinkjoseph, line 4: I see that you split the files here and wrote the functions in separate files each representing the sub module.
However, I think that these have to be defined as functions in a file called `functions.ts` in the same directory as ducks module, that is as `src/ducks/module/functions.ts`
The reason being, these functions are required only in the action creator and we won't be interacting with the data sources directly.
We will only interact with the data sources through redux using the action creators.
If a function really is being used again in another ducks module (which I think is rare), it should be put in a `functions.ts` that is common to all the ducks module at `src/ducks/functions.ts` but still not at a http or firebase resources directory that is outside the ducks directory.
Also, I feel that it makes more sense to put the helpers related to making `HTTP` requests in `src/http/resources` rather than the actual `http` or `firebase` operations that are to be carried out.
And on this note, only helpers that aid `HTTP` requests should be put in this directory.
helpers aiding `firebase` should be in `src/firebase/resources.`
@jpmarques66

web-client/src/ducks/auth/phone/types.ts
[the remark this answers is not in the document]
ashwinkjoseph: We had a discussion over this in PM.
We are considering various options on how to do this. but we decided to raise this as a separate issue for now

web-client/src/ducks/auth/phone/functions.ts
```diff
@@ -0,0 +1,23 @@
+import { IOTPAuth, IPhoneNumberAuth } from 'src/ducks/auth/phone/types';
```
ashwinkjoseph, line 1: we could use ./types right? but it achieves the same thing

web-client/src/ducks/auth/reducer.ts
```diff
@@ -1,103 +1,20 @@
...
-    },
-    [FIREBASE_PHONE_VERIFY.REJECTED]: (
-      state: AuthState,
-      { payload }: { payload: Error },
-    ) => {
-      state.error = payload;
-      state.confirmationResult = undefined;
-      state.loading = false;
-    },
+  merge(facebookReducer, phoneReducer, {
```
ashwinkjoseph, line 17: do we need to use the merge function? can't we just use the spread operator? is there any issue that using spread operator can cause here that merge will solve?

web-client/src/ducks/auth/types.ts
```diff
@@ -1,46 +1,20 @@
 import createActionTypeFactory from 'src/store/utils/createActionTypeFactory';
 
-const { asyncType } = createActionTypeFactory('AUTH');
-const { asyncType: firebaseAsyncType, observerType } = createActionTypeFactory(
-  'FIREBASE',
-);
+export const { asyncType } = createActionTypeFactory('AUTH');
+export const { observerType } = createActionTypeFactory('FIREBASE');
```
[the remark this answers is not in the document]
ashwinkjoseph: This was corrected

web-client/src/ducks/auth/phone/actions.ts
ashwinkjoseph: This was corrected

web-client/src/ducks/auth/facebook/actions.ts
ashwinkjoseph: This was corrected

web-client/src/ducks/auth/facebook/types.ts
ashwinkjoseph: This was corrected

web-client/src/ducks/auth/phone/actions.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import { IOTPAuth, IPhoneNumberAuth } from '../types';
```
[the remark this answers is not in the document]
ashwinkjoseph: We had a conversation where we decided to create new issue regarding deciding the best standard for defining the interfaces.
So decision on this will be taken in that issue and will be outside of this PR

web-client/src/ducks/auth/reducer.ts
jpmarques66: Agree.

web-client/src/ducks/auth/phone/functions.ts
jpmarques66: auto imports 😄

ashwinkjoseph: oh okay 😅

PR #390, opened 2020-04-13
fix: updated some of the code

sabind:
## Description

Update some of the trigger functions

## Motivation

Reviewing the code @ashwinkjoseph wrote + adding some changes.

## Release Checklist

## Security Checklist

### DIFF
--- a/functions/src/offers/index.ts
+++ b/functions/src/offers/index.ts
@@ -1,5 +1,4 @@
 import { validateOrReject } from 'class-validator';
-import { firestore } from 'firebase';
 import * as admin from 'firebase-admin';
 import * as functions from 'firebase-functions';
 import { Change, EventContext } from 'firebase-functions/lib/cloud-functions';
@@ -24,21 +23,17 @@ const queueStatusUpdateTriggers = async (
     offerBefore?.status !== OfferStatus.accepted &&
     offerAfter?.status === OfferStatus.accepted
   ) {
+    // Update Request with new status and CAV.
     operations.push(offerAfter.requestRef.update({
       cavUserRef: offerAfter.cavUserRef,
-      status: RequestStatus.ongoing
+      status: RequestStatus.ongoing,
     }));
-    let offersToReject = await db.collection('offers').where('requestRef', '==', offerAfter.requestRef).get();
-    offersToReject.forEach(offer => {
-      if(offer.ref !== change.after.ref){
-        operations.push((async () => {
-          await offer.ref.update({
-            status: OfferStatus.rejected
-          });
-          Promise.resolve();
-        })()); 
-      }
-    });
+
+    // TODO: Notify CAV that they have been selected.
+
+    // We won't update all offers because it's a waste of reads + writes + additional cloud function triggers
+    // This is not done in a safe manner -- we should use transactions and batched writes, but also remembering
+    // that batched writes are limited to 500 operations so we need to chunk these operations.
   }
 
   return Promise.all(operations);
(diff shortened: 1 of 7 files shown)


functions/src/offers/index.ts
```diff
@@ -24,21 +23,17 @@ const queueStatusUpdateTriggers = async (
...
-    offersToReject.forEach(offer => {
-      if(offer.ref !== change.after.ref){
-        operations.push((async () => {
-          await offer.ref.update({
-            status: OfferStatus.rejected
-          });
-          Promise.resolve();
-        })()); 
-      }
-    });
```
sabind, line 41: We didn't need to do this as it would be quite expensive and we don't need to `reject` those offers, particularly if the first one falls through.

functions/src/offers/index.ts
```diff
@@ -55,7 +50,7 @@ export const triggerEventsWhenOfferIsCreated = functions.firestore
   .onCreate((snapshot: DocumentSnapshot, context: EventContext) => {
     return validateOffer(snapshot.data() as IOffer).catch(errors => {
       console.error('Invalid Offer Found: ', errors);
-      return firestore()
+      return db
```
sabind, line 53: This happens all over the code (a mistake I made initially). We should make sure the cloud functions use `admin` db.

functions/src/requests/index.ts
```diff
@@ -1,125 +1,147 @@
...
   if (
     requestBefore?.status !== RequestStatus.completed &&
     requestAfter?.status === RequestStatus.completed &&
     requestAfter.cavUserRef
   ) {
-    let user = UserFirestoreConverter.fromFirestore((await requestAfter.cavUserRef.get()));
+    // Using increment without read
     operations.push(requestAfter.cavUserRef.update({
-      casesCompleted: user.casesCompleted + 1
+      casesCompleted: admin.firestore.FieldValue.increment(1),
```
sabind, line 66: An interesting firebase feature that lets us increment values without having to consume a read. It also works in a much safer fashion. `read before write` operations like the one this replaces are unsafe (unless conducted inside a transaction).

functions/src/users/index.ts
```diff
@@ -13,15 +16,15 @@ const validateUser = (value: IUser): Promise<void> => {
...
-        .delete();
-    });
+    return validateUser(snapshot.data() as IUser)
+      .catch(errors => {
+        console.error('Invalid User Found: ', errors);
+        return db
+          .collection('users')
+          .doc(context.params.userId)
+          .delete();
+      });
```
sabind, line 27: just a formatting thing, but we should always put the `.then` `.catch` or `.finally` on new lines so the reader can see which promise resolution this block is for.

functions/src/offers/index.ts
ashwinkjoseph: okay. I looked at the status Object and saw that there was a `rejected` status.
So I assumed that the requirement was to invalidate all offers and only keep one as accepted.
This makes sense

functions/src/offers/index.ts
ashwinkjoseph: okay cool.
I was using db whenever I needed to perform an operation on any other document.
Just left these untouched.
I will fix it in other parts of the code as well

functions/src/requests/index.ts
ashwinkjoseph: oh. this is great. Thanks!

functions/src/users/index.ts
ashwinkjoseph: okay

Issue #290, opened 2020-04-07
Add basic pageview analytics to the map


s0:
We want to be able to track basic pageviews (not unique users) before dropping any cookies, and probably what country they're from (to help us focus on audiences that are using the map). This will allow us to avoid requiring user consent in the EU to drop a cookie simply track pageviews.

We'll need some backend service for this, or a 3rd party provider.

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/290)

*This information is updated automatically. To modify it, please use ZenHub.*

PR #401, opened 2020-04-14
Added Google Analytics for the map only.

kfo42:
# Description

Addresses issue #290 

## Release Checklist

## Security Checklist

### DIFF
--- a/map/public/index.html
+++ b/map/public/index.html
@@ -1,6 +1,16 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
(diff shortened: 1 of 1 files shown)


sabind:
We should use Firebase for this so this will have to change.

We also have to be quite careful when adding this even just to the map. @s0 or @mdeous knows more.

sabind:
That being said it would be possible to add a cookie banner that complies with GDPR which would be a huge help.

image

sabind:
The above being said we can use Firebase to track the analytics and react to them. We really could use the cookie banner, even if it’s super painful to do just to collect analytics.

PR #406, opened 2020-04-14
feat(map): allow users to submit info

### LINKED ISSUE #286: Allow users to submit a new marker by drawing directly on the map and filling in info
The most difficult / time-consuming part of adding new datapoints to the map is turning location names into lat/long + radius. (The current process we've been using is to use google maps to do this, documented here: https://github.com/reach4help/reach4help/tree/master/map/src/data), the second most time consuming thing is converting the contact information into the right format and types.

Before the map data can be fully crowd-sourced, and to decrease our own workload before then, we need to improve this.

It would be great if users could directly draw on the map to indicate lat / long, and radius, and then be prompted to fill in the required information (location name, group name, group description, and contact details), and then have the json directly created for us which we can approve later.

I envisage the workflow going something like this:

1. User clicks "Add information to this map"
2. Map displays an overlay that says something like "move the map to include the area served", with a button to "continue". (the overlay could have a style like `pointer-events: none;` to allow the user to interact with the map underneath.
3. After a user clicks continue, Map then displays "click the map to indicate the center of the area served"
4. At this point, the we track the user's cursor to calculate the radius of the area, and say "move your cursor to the desired radius, and then click"
5. Then we say "Is this area okay?", with buttons "Redo" and "Yes, Continue".
6. We then display a form for the user to fill out with the remaining information and a submit button.

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/286)

**Belonging to Epics:**

* **[EPIC]** [[@isma-osy](https://github.com/isma-osy)] [#413 - Prepare MAP for Go-To-Market / Marketing strategy](https://github.com/reach4help/reach4help/issues/413)

**Blocked By:**

* ~~[#286 - [MAP] Rework model for markers](https://github.com/reach4help/reach4help/issues/414)~~

btuerker:
## Description
Hi there!

I'm going to write down the problematic parts I noticed. 

1. First things first, I changed almost the entire AddInstruction component. So, maybe we should call it AddInformation or something like?
2. We can refactor some of the state variable names added recently in the App component
a.addInfoStep -> currentAddInformationStep
b.setAddInfoStep -> setCurrentAddInformationStep
3. We can make an improvement on the set-marker step. Simply we can allow a user to create a marker by a simple click event on the map. I tried and it works but caused some problems at removing that listener as well as requires to check is there a listener when we step back. 
3. I've added a table view of the final state in AddInstruction render. It's absolutely unnecessary. I added it just to see everything is fine. During this review, I'll remove it.
4. I'm not good at English :) So, I need help for defining better heading texts.
5. We can make Add Instruction button on the header invisible or we can alter that with reverse action so it closes add instruction panel.
6. During the add information process, we can hide the other markers, clusters, and circles. (I solved this in a dream(yes I'm coding while dreaming) yesterday 😄 We can just add another filter such as 'none' or 'clear', then ba dum tss)
7. Inside the AddInstruction render method, we can extract steps and entire form as a separate component. That will make it better.
8. I used a Map object to manage service checkboxes, there would be a better way...
## Release Checklist

## Security Checklist

## Additional Notes

Screen Shot 2020-04-14 at 9 49 16 PM
Screen Shot 2020-04-14 at 9 49 20 PM
Screen Shot 2020-04-14 at 9 49 41 PM
Screen Shot 2020-04-14 at 9 50 02 PM
Screen Shot 2020-04-14 at 9 50 07 PM

closes #286


map/src/components/add-instructions.tsx
```diff
@@ -1,70 +1,582 @@
...
+                <thead>
+                  <tr>
+                    <td>variable</td>
+                    <td>value</td>
+                  </tr>
+                </thead>
+                <tbody>
+                  <tr>
+                    <td>center</td>
+                    <td>{JSON.stringify(center)}</td>
```
micaelr95, line 534: I don't think is very good idea to use json stringify to show data to the user.

I'ts safer to put `lat: {center.lat}, long: {center.lng}`

map/src/components/add-instructions.tsx
```diff
@@ -1,70 +1,582 @@
...
+                    <td>form - name</td>
+                    <td>{form.name}</td>
+                  </tr>
+                  <tr>
+                    <td>form - area</td>
+                    <td>{form.area}</td>
+                  </tr>
+                  <tr>
+                    <td>form - services</td>
+                    <td>{JSON.stringify([...form.services])}</td>
```
micaelr95, line 550: Same as last comment

map/src/App.tsx
```diff
@@ -250,8 +268,8 @@ export default styled(App)`
         position: absolute;
         z-index: 100;
         max-width: 500px;
-        top: 10px;
-        left: 10px;
+        top: ${p => p.theme.spacingPx}px;
+        left: ${p => p.theme.spacingPx}px;
```
micaelr95, line 272: Can you explain this function?

Can't you just put `theme.spacingPx` or something similar?

sabind:
I’d recommend looking at #411 and confirm the data collected matches the model. This also means using Firebase to persist the map marker. You may use the SDK to add data, but it must be the correct format.

s0:
I'll take ownership of getting this form storing data in firebase, before that we need to decide on #414 though, i'll work on that now.

map/src/App.tsx
s0: no, p.theme is provided by the styled-components theme provider, p is the properties for the components.

s0: See: https://styled-components.com/docs/advanced

micaelr95: I've never used styled components too much, just to do simple things.

Thanks for the explanation :smile:

map/src/components/add-instructions.tsx
btuerker: Yes, it's right! I've put them just for experimentation purposes. I have no idea yet for the last step screen. Just planning to put some information messages such as "Thank you!  We'll review your application soon". Maybe we can put the data posted but not that JSON way 😄

micaelr95: I think just the message you said is fine for now. If really necessary then we can show a table with the inserted data like you have in the screenshot but more formatted.

sabind:
#418 is approved to merge.

map/src/components/add-instructions.tsx
```diff
@@ -1,70 +1,582 @@
...
+      radius: null,
+      form: {
+        name: '',
+        area: '',
+        services: Object.keys(SERVICES).reduce(
+          (map: Map<string, boolean>, key: string) => {
+            map.set(key, false);
+            return map;
+          },
+          new Map(),
```
s0, line 70: Nice instantiation btw, i like this, may use this in the future!

s0:
> I used a Map object to manage service checkboxes, there would be a better way...

Regarding this comment too, i agree yes, probably just a `Set` of the services that are enabled rather than a map to boolean objects.

s0:
This is basically ready now. Just need to deploy #469 and set up the environment variables in netlify for the appropriate application.

I think it probably makes sense to use production for now, given that we're actually "launching" this, cc @sabind thoughts?

PR #407, opened 2020-04-14
Completed the TODOs in firestore triggers

ashwinkjoseph:
## Description
Completed all the TODO items in firestore triggers which include:

- Updating the CAV reference in the request when an offer is accepted
- Increasing `casesCompleted` for a CAV user when the a request assigned to the CAV user is completed
- Adjusting the CAV user's rating based on new rating
- Adjusting the PIN user's rating based on new rating
- Adjusting the CAV user's rating when the rating for an existing request is changed
- Adjusting the PIN user's rating when the rating for an existing request is changed
- Sending notification to a CAV user when the CAV user's offer is accepted for a request

## Release Checklist

## Security Checklist


functions/src/offers/functions.ts
```diff
@@ -0,0 +1,78 @@
...
+    operations.push(
+      offerAfter.requestRef.update({
+        cavUserRef: offerAfter.cavUserRef,
+        status: RequestStatus.ongoing,
+      }),
+    );
+    
+    messaging.send({
+        data:{
+            entity: 'offer',
```
sabind, line 31: if possible we should send data for the following:
- Render a "toast" notification on the front end which includes maybe sending some information about the offer
- Compose a text message with enough info for the user to act on it.

Ideally there would be a pub/sub cloud function that allows anyone to add messages to that get converted into SMS messages to the client (if they have `text me updates`) turned on.

This is down the road, but we should likely put a little more detail into the message body for these purposes.

ashwinkjoseph: I was thinking of that but then I thought it was better to fetch that data from the device once the notification arrives as I had given the details of the entity in a generic form.
but Yes we can avoid additional reads with this method.
I will make the change

ashwinkjoseph: also regarding twilio, I was under the impression that twilio was supposed to be done only through cloud functions.
We will expose a cloud function with HTTP invocation and look for authentication of users in it.

sabind: Yes twilio should only be done from cloud functions (serverside), but I think hooking a cloud function trigger to the pub/sub topic allows us to concentrate the notification logic into the pub/sub topic for each user. This way anyone in the larger product can notify a user via SMS if publish to their notification topic vs having an HTTP endpoint. Simultaneously it creates a singular place to check the users own notification preferences.

ashwinkjoseph: okay. that makes sense.

ashwinkjoseph: I am making the modification now

ashwinkjoseph: @sabind I tried looking into the parameters. but for the offer, the only attribute that can be used by the user to understand what the offer is about is the `message` attribute.
However, this may not be very meaningful for notification as the `message` may be out of context without the details about the `request` for which the offer was made.
We do have meaningful attributes in the `request`. But if we need to send them, we would need to fetch the details of the `request` which is an additional read on the trigger.
If we are doing another additional read, I feel it's better to do it on the `client-side` so as to be able to show the `request` as well if the user were to click on the notification.
However, if your goal is to add it for the purpose of the SMS, then it makes sense and I can read from the `request` object as well

sabind: I'd simply say: " You have new offer to an open request: https://reach4help....<request_id>" and include the message.

ashwinkjoseph: okay I have made the change for this and have pushed it now

PR #419, opened 2020-04-17
Map/#76

rim23ma:
## Description

Add localization/internationalization to Map project
1. Added drop down to select language;
2. If language selected then local storage set "selected language"
3. If there is no "selected language" in a local storage, then check if query has language preference, then check if browser has language option, if not then set it ti default "en"
4. Put some russian translation

Screen Shot 2020-04-16 at 9 41 51 PM

Screen Shot 2020-04-16 at 9 42 02 PM

## Release Checklist

## Security Checklist


map/src/components/header.tsx
```diff
@@ -74,8 +81,8 @@ const Header = (props: Props) => {
         >
           <MdAdd className="icon icon-left" />
           <span>
-            Add information
-            <span className={CLS_SCREEN_LG_ONLY}>&nbsp;to this map</span>
+            {t(s => s.mdAdd1)}
+            <span className={CLS_SCREEN_LG_ONLY}>&nbsp;{t(s => s.mdAdd2)}</span>
```
s0, line 85: Rather than have the translations split like this, i think it makes more sense to provide both a short and long string, and then switch between the two (splitting like this may not be possible in all languages), and will be complicated for RTL languages for sure.

map/src/data/index.ts
```diff
@@ -45,3 +47,23 @@ export const SERVICE_TYPES = Object.keys(SERVICES) as Service[];
...
   service?: Service;
 }
+
+export const LANGUAGES = {
+  en: {
+    label: t(s => s.languages.en),
+  },
+  ru: {
+    label: t(s => s.languages.ru),
+  },
```
s0, line 57: I think rather than providing language names in the current language, it's better to display languages in the *target* language, so translations for language names are unnecessary.

map/src/data/index.ts
```diff
@@ -1,36 +1,38 @@
...
     color: COLORS.orange,
   },
   financial: {
-    label: 'Financial',
+    label: t(s => s.services.financial),
     color: COLORS.orange,
   },
   information: {
-    label: 'Information',
+    label: t(s => s.services.information),
```
s0, line 35: setting the labels in the constant like this means that when the language changes, the labels themselves won't change unless the page is refreshed. Instead we can use the keys of the service to look up the label in the translation when displaying it.

map/src/data/index.ts
```diff
@@ -45,3 +47,23 @@ export const SERVICE_TYPES = Object.keys(SERVICES) as Service[];
...
+export type Language = keyof typeof LANGUAGES;
+
+export const isLanguage = (language: string): language is Language =>
+  language in LANGUAGES;
+
+export const LANGUAGE_TYPES = Object.keys(LANGUAGES) as Language[];
+
+export interface Translate {
+  language?: Language;
+}
```
s0, line 69: This information feels like it's mostly duplicated from `i18n/index.ts`, and would better live there. The `data` directory is more for data we're displaying, not UI implementation (which `i18n` would come under).

map/src/i18n/index.ts
```diff
@@ -0,0 +1,35 @@
...
+  ru,
+};
+
+type Language = keyof typeof STRINGS;
+
+const selectedLanguage = localStorage.getItem('selectedLanguage');
+
+const queryLanguage = window.location.href.split('/')[3];
+
+const browserLanguage = window.navigator.language.split('-')[0];
```
s0, line 16: written like this, it means that the language will only be fetched / calculated once, upon browser load. A page refresh shouldn't be required to change the language.

map/src/data/index.ts
```diff
@@ -45,3 +47,23 @@ export const SERVICE_TYPES = Object.keys(SERVICES) as Service[];
...
+};
+
+export type Language = keyof typeof LANGUAGES;
+
+export const isLanguage = (language: string): language is Language =>
+  language in LANGUAGES;
+
+export const LANGUAGE_TYPES = Object.keys(LANGUAGES) as Language[];
+
+export interface Translate {
```
s0, line 67: I think it would be better to name this `Internationalization`

map/src/i18n/index.ts
```diff
@@ -0,0 +1,35 @@
...
+import ru from './langs/ru';
+
+const STRINGS = {
+  en,
+  ru,
+};
+
+type Language = keyof typeof STRINGS;
+
+const selectedLanguage = localStorage.getItem('selectedLanguage');
```
s0, line 12: the storage key should be hardcoded

map/src/i18n/index.ts
```diff
@@ -0,0 +1,35 @@
...
+
+const queryLanguage = window.location.href.split('/')[3];
```
s0, line 14: This is not a good / reliable way to get the first path component of the URL. In addition, it's a bit less flexible than using the query string, as the variable indicates it is doing.

map/src/App.tsx
```diff
@@ -52,9 +54,17 @@ class App extends React.Component<Props, State> {
...
       resultsMode: 'open',
       updateResultsOnNextClustering: false,
+      translate: {},
     };
   }
 
+  componentDidUpdate(prevProps: Props, prevState: State): void {
+    const { translate } = this.state;
+    if (prevState.translate !== translate) {
+      window.location.reload(false);
```
s0, line 64: we can avoid reloading the page when the language updates by handling state a little differently

s0: (for future reference, this could also have been put in the `setTranslate` function directly)

map/src/App.tsx
```diff
@@ -31,6 +32,7 @@ interface State {
   addInstructionsOpen: boolean;
   fullScreen: boolean;
   updateResultsOnNextClustering: boolean;
+  translate: Translate;
```
s0, line 35: A good way to manage this state would be to add event listeners in i18n that can update this state to filter changes down through the component tree. That would also allow things like a language changing in one window for the map also changing in all other tabs where the map is open.

map/src/App.tsx
```diff
@@ -120,6 +130,12 @@ class App extends React.Component<Props, State> {
     }));
   };
 
+  private setTranslate = (mutator: TranslateMutator) => {
+    this.setState(state => ({
+      translate: mutator(state.translate),
+    }));
+  };
+
```
s0, line 138: rather than providing a callback that allows components to change the language in the App component state, it would be better to make these requests go via i18n, and trigger an event to update the state in `App`

map/src/components/header.tsx
```diff
@@ -33,24 +39,16 @@ const Header = (props: Props) => {
...
-            </p>
+            <h1>{t(s => s.title)} - Reach4Help</h1>
+            <p>{t(s => s.info)}</p>
             <p className="muted">
-              This map is part of&nbsp;
-              <a href="https://reach4help.org">reach4help.org</a>, a
-              volunteer-run project. It is open source and can be&nbsp;
+              {t(s => s.source1)}&nbsp;
+              <a href="https://reach4help.org">reach4help.org</a>
+              {t(s => s.source2)}&nbsp;
```
s0, line 47: probably not necessary for this PR, but we could do with using placeholders, or allowing template functions with HTML escaping instead of having to split translation around html components like this.

Placeholders could be achievable in this PR though.

map/src/i18n/index.ts
s0: Also, this will only check the first language of the navigator, and not check to see if there's perhaps a secondary language that would be better. we should check all languages and filter those we don't know

PR #425, opened 2020-04-19
Additional Field in Privileged User Info

ashwinkjoseph:
## Description
This PR adds an additional data field to the Privileged Information Sub Collection of a User Document that stores the "raw" address entered by the user apart from the one obtained from Google Geocode API.
This PR also completes the onCreate trigger for the privileged information sub-collection.
## Motivation
Users should be able to see the exact data that is being entered by them.
We should use Geocode API to retrieve the Latitude and Longitude of the address entered by the user.
The current Interface suggested storing the result of the Geocode operation directly in the document as the address.
The problem with this is that the address fields returned by Google may not be the exact ones that the user has agreed as correct by typing it, even if it is close to the one that the user entered or more precise than the one that the user has entered.
It is a better option to record the two separately and keep a record of both rather than just one.
This way, we can have a record of exactly what the user typed in and agreed as the correct address and also the precise (accuracy is debatable as it depends on the region, for example, the response in my region is not very accurate) address that Google Geocode API returns

## Release Checklist

## Security Checklist

### DIFF
--- a/functions/src/models/users/privilegedInformation.ts
+++ b/functions/src/models/users/privilegedInformation.ts
@@ -13,11 +13,12 @@ interface IUserAddress {
   city?: string;
   state?: string;
   country?: string;
+  coords?: firebase.firestore.GeoPoint;
 }
 
 export interface IPrivilegedUserInformation extends DocumentData {
   addressFromGoogle: google.maps.GeocoderResult;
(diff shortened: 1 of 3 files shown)


sabind:
LGTM - Good call on recording both (for many good reasons) 👍

PR #446, opened 2020-04-22
Refactored Notifications and Added Twilio SMS support

ashwinkjoseph:
## Description
This PR refactors the way the notifications are sent to use a dispatch module which decides the methods that are to be used for notifying based on the configuration passed onto it.
This PR also adds SMS support using Twilio.
## Motivation
We needed to notify users with SMS as well as Push Notifications as not everyone might get push notifications.
At the same time, we wanted to abstract the choice of communication away from the actual trigger logic
## Release Checklist

## Security Checklist


functions/src/app.ts
```diff
@@ -0,0 +1,4 @@
+import * as admin from 'firebase-admin';
+
+// Do this only once at the top level.
+export const app = admin.initializeApp();
```
sabind, line 4: Added this to make sure we init `app` only once and can re-use everywhere. Running the init in multiple files creates an error on deploy.

functions/src/dispatch/fcm.ts
```diff
@@ -0,0 +1,6 @@
+import * as Admin from 'firebase-admin';
+import { app } from '../app';
+
+const messaging = app.messaging();
+
+export const sendToTopic = async (config: Admin.messaging.Message) => messaging.send(config);
```
sabind, line 6: Using the imported `app` to access `messaging()`

functions/src/twilio/index.ts
```diff
@@ -0,0 +1,15 @@
+import * as functions from 'firebase-functions';
+import * as Client from 'twilio';
+
+const TWILIO_ACCOUNT_SID = functions.config().twilio.account_sid;
+const TWILIO_SERVICE_SID = functions.config().twilio.service_sid;
+const TWILIO_AUTH_TOKEN = functions.config().twilio.auth_token;
```
sabind, line 6: Using the firebase config service here.

sabind:
Added some edits and ran the linter/prettier to test if it would deploy

image

functions/src/app.ts
ashwinkjoseph: But that's why I initialized different instances with different names for each instance.
I am not sure about deployment, but the error during runtime wouldn't exist with different names.
I wanted to keep each module independent as much as possible

sabind: I don't think that's necessary -- different modules with different instances.

sabind: Each module runs in a different instance regardless as they all run in different cloud functions.

ashwinkjoseph: okay, that works too. but should we also make this change for the cloud function files for each model as right now we are initializing the admin SDK in every file.

we could implement it like keep the initialization in one single file that exports the initialized Admin SDK instance, just as it is currently being done in the web-client app.

ashwinkjoseph: do you want me to make this change @sabind ?

sabind: I think I made the changes everywhere necessary. I will revisit this one we merge this and the testing PR in.

sabind:
@ashwinkjoseph I resolved the conflicts and moved the *are we testing...* initialization to the `app.ts` file. `yarn test` automatically sets `NODE_ENV=test` so the code now is handling this automatically. When the code runs in firebase `NODE_ENV=production` so it inits the app properly.

sabind:
@ashwinkjoseph - If you agree with the changes, feel free to merge.

PR #469, opened 2020-04-26
feat(backend): allow anyone to add markers to datastore

s0:
## Description

Ahead of #406

Allows anyone to write to create new markers (not edit existing ones), that save directly to the firestore database.

## Motivation

To be able to accept user data using the new form, allowing users to write directly to the firestore when creating new markers seemed like the best response.

This data isn't currently being used / loaded in the frontend. After this gets in and the initial form gets deployed, I'm going to add a new boolean property called `visible` that (can't be true when users submit) that will be indexed on, to allow users to query for the visible markers only. Marker submissions will then (for now) require a review by someone with appropriate access (currently only admins).

Beyond this I plan on allowing users to optionally view "hidden" fields before adding new information so that they can ensure they're not causing duplicate information.

Also, I want to create a function that runs whenever a marker's data is updated to keep a log of a marker's history.

cc @sabind @lenjoseph thoughts?

## Release Checklist

## Security Checklist

cc @reach4help/security

## Additional Notes

We probably want to have unit tests for these rules too, but I was unable to really make progress as the tests aren't currently working locally for me (we should also update the workflows to make sure that they run in CI too, and don't become broken).

### DIFF
--- a/firebase/firestore.rules
+++ b/firebase/firestore.rules
@@ -1,5 +1,17 @@
 rules_version = '2';
 
+function verifyMarkerDataComplete(data) {
+   return ('contentTitle' in data) &&
+          ('type' in data) &&
+          ('type' in data.type) &&
+          ('contact' in data) &&
+          ('loc' in data) &&
+          ('description' in data.loc) &&
+          ('lat' in data.loc) &&
+          ('lng' in data.loc) &&
+          ('serviceRadius' in data.loc)
+}
+
 function isAuthenticated() {
   return request.auth.uid != null;
 }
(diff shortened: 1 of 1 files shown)


firebase/firestore.rules
```diff
@@ -1,5 +1,17 @@
 rules_version = '2';
 
+function verifyMarkerDataComplete(data) {
+   return ('contentTitle' in data) &&
+          ('type' in data) &&
+          ('type' in data.type) &&
+          ('contact' in data) &&
+          ('loc' in data) &&
+          ('description' in data.loc) &&
+          ('lat' in data.loc) &&
```
ashwinkjoseph, line 10: `lat` and `lng` should combined into a separate field in the document of the GeoPoint data type.
This is advised as we can perform geospatial queries later on.
for validation of the data type, we have already written a function with the name `verifyLatLngObject` within this `firestore.rules` file

s0:
@ashwinkjoseph done, and also added a condition for the visible property to be there (and specified as false)

Issue #385, opened 2020-04-11
[App shell] - Add top-bar and side bar


jpmarques66:
**Is your feature request related to a problem? Please describe.**
Implement top and sibe-bar.

**Describe the solution you'd like**
Create a container that receives children and renders them inside a top and side-bar, like the following image.
![image](https://user-images.githubusercontent.com/52417176/79050200-64ca5200-7c20-11ea-9889-675c40980290.png)

**Additional context**

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/385)

*This information is updated automatically. To modify it, please use ZenHub.*

**Belonging to Epics:**

* **[EPIC]** [#322 - Core Components](https://github.com/reach4help/reach4help/issues/322)

PR #481, opened 2020-04-27
feat(web-client): dashboard layout

### LINKED ISSUE #321: Component: Layout / Header + Side Drawer
--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/321)

**Belonging to Epics:**

* **[EPIC]** [#322 - Core Components](https://github.com/reach4help/reach4help/issues/322)

**Children:**

* [x] ~~[#323 - design component: layout / header + side drawer](https://github.com/reach4help/reach4help/issues/323)~~
* [ ] **[BLOCKING: [#102](https://github.com/reach4help/reach4help/issues/102), [#160](https://github.com/reach4help/reach4help/issues/160), [#162](https://github.com/reach4help/reach4help/issues/162), [#348](https://github.com/reach4help/reach4help/issues/348)]** [[@btuerker](https://github.com/btuerker)] [#324 - implement component: layout / header + side drawer](https://github.com/reach4help/reach4help/issues/324)

btuerker:
## Description
Hi there!
I've just completed the layout component.

I'm going to comment on the things I couldn't resolve and I require assistance in the below comments. 
## Release Checklist

## Security Checklist

## Additional Notes

closes #321 #323

Visual changes:
Screen Shot 2020-04-27 at 11 58 18 PM
Screen Shot 2020-04-28 at 12 14 22 AM
Screen Shot 2020-05-05 at 5 25 15 PM


web-client/src/components/SideNavMenu/SideNavMenu.tsx
```diff
@@ -0,0 +1,130 @@
...
+  return (
+    <>
+      {item.location ? (
+        // TODO I couldn't find a proper way here to solve type errors
+        // TODO Location is a JS library
+        // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+        // @ts-ignore
+        <Link to={{ pathname: item.location.path }} onClick={closeSider}>
+          {menuItem}
+        </Link>
```
btuerker, line 54: Any advice here rather than @ts-ignore?

web-client/src/modules/request/index.ts
```diff
@@ -1,10 +1,92 @@
...
 import { Module } from 'src/types/module';
 
+import { MenuItem } from '../../types/menu-item';
 import ContentPage from './pages/ContentPage';
+import { NewRequestLocation } from './pages/routes/NewRequestRoute/constants';
+
+// TODO (remove link id): ask help to find a better way to keep track of ids
+export const menuItems: Array<MenuItem> = [
+  {
+    id: uniqueId(),
```
btuerker, line 21: This is a crucial thing needs to solved by another approach. I'm passing these unique ids to Antd Menu component as a key prop. It uses these keys to find which one is active, clicked, etc. From now, I didn't mark the default selected key on initialization and it's a big problem. I've just an idea to use browser location i.e 'requests/open' as id for menu item whic has a location property which means it's a link. Let's say a user is in '/requests/new' location so we can make it active-selected.

web-client/src/modules/request/index.ts
```diff
@@ -1,10 +1,92 @@
...
+    icon: React.createElement(FileAddOutlined),
```
btuerker, line 22: I'm not sure that this is a good way. I'm open for any advice in here.

web-client/src/modules/request/index.ts
```diff
@@ -1,10 +1,92 @@
...
+    ],
+  },
+];
 
 const requestModule: Module = {
   path: '/requests',
   component: ContentPage,
   protected: true,
+  layout: 'dashboard',
+  menuItems,
```
btuerker, line 90: As we discussed with @ashwinkjoseph this is not required at all because most of our pages will use the same layout except login steps. But I couldn't figure out how to merge DashboardRoute and ProtectedRoute at this moment. That's why I used this for just experimentation. Maybe the DashboardRoute is unnecessary, then we can just use only the ProtectedRoute.

web-client/src/pages/MasterPage.tsx
```diff
@@ -8,13 +8,25 @@ import {
...
+      if (routeModule.layout === 'dashboard') {
+        return (
+          <DashboardRoute
+            key={moduleName}
+            path={routeModule.path}
+            component={routeModule.component}
+            menuItems={routeModule.menuItems || []}
+          />
+        );
+      }
```
btuerker, line 28: Same as above

web-client/src/pages/routes/DashboardRoute.tsx
```diff
@@ -0,0 +1,43 @@
...
+import DashboardLayout from '../../components/DashboardLayout/DashboardLayout';
+
+// TODO remove when using real profile state
+const mockProfileData = {
+  fullName: 'John Watkins',
+  photoUrl:
+    'https://lh3.googleusercontent.com/-6DkdLho7c5Y/AAAAAAAAAAI/AAAAAAAAAAA/AMZuuckYv30Z33APqrrucaWdEBbadI7WxQ/photo.jpg?sz=46',
+  followers: 15,
+  stars: 4.5,
+};
```
btuerker, line 15: I think here we'll fetch the profile state from redux rather than this mock. I've just checked the user model but left it to do later by a piece of advice.

web-client/src/pages/routes/DashboardRoute.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+// TODO change locations
+const siteLocations = {
+  logout: new Location('/requests/logout'),
+  contact: new Location('/requests/contact'),
+  map: new Location('/requests/map'),
+  notifications: new Location('requests/notifications'),
+};
```
btuerker, line 23: I've some ideas but it'll be better to follow some advice. As I see, the menuLinks and these siteLocations will be truly static, maybe we can find a way to make it only object like `siteLinks={ menuLinks: { profile: {logout: {location: '/logout'}}, request: {...}} general: {contact:{ location:'/contact'}. etc }}`.

web-client/src/pages/routes/DashboardRoute.tsx
```diff
@@ -0,0 +1,43 @@
...
+const DashboardRoute: React.FC<DashboardRouteProps> = ({
+  path,
+  component,
+  menuItems,
+}) => (
+  <DashboardLayout
+    menuItems={menuItems}
+    profileData={mockProfileData}
+    siteLocations={siteLocations}
+  >
```
btuerker, line 34: This is the actual prop provider. These props are passing through their children and so on. I'm not so experienced and didn't use it but using Context API can solve the prop passing problem.

web-client/src/modules/request/index.ts
```diff
@@ -1,10 +1,92 @@
...
+import React from 'react';
+import Location from 'react-app-location';
 import { Module } from 'src/types/module';
 
+import { MenuItem } from '../../types/menu-item';
 import ContentPage from './pages/ContentPage';
+import { NewRequestLocation } from './pages/routes/NewRequestRoute/constants';
+
+// TODO (remove link id): ask help to find a better way to keep track of ids
+export const menuItems: Array<MenuItem> = [
```
btuerker, line 19: @ashwinkjoseph This is the only resource for recursive menu generation. But, I don't think that it's the right place. Any advice about where to store this static data?

web-client/src/types/menu-item.d.ts
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import Location from 'react-app-location';
+
+export interface MenuItem {
```
btuerker, line 4: Is that right to declare type in here as it's only used by SideNavMenu? Maybe we should export this inside the SideNavMenu.tsx

web-client/src/components/DashboardLayout/DashboardLayout.tsx
```diff
@@ -0,0 +1,49 @@
...
+interface DashboardLayoutProps {
+  menuItems: Array<MenuItem>;
+  profileData: {
+    fullName: string;
+    photoUrl: string;
+    followers: number;
+    stars: number;
+  };
+  siteLocations: any;
+  children?: React.ReactNode;
```
btuerker, line 46: This is like a duplicate code. But, I really don't know how to fix it. Any advice? Can Context API solve it?

web-client/src/components/SideDrawer/SideDrawer.tsx
```diff
@@ -0,0 +1,138 @@
...
+            <TeamOutlined />
+            {profileData.followers}
+          </ProfileDetail>
+          <ProfileDetail>
+            <StarOutlined />
+            {profileData.stars}
+          </ProfileDetail>
+        </ProfileDetails>
+      </ProfileContent>
+    </ProfileWrapper>
```
btuerker, line 45: This can be extracted to its own component

web-client/src/modules/request/index.ts
puzzledbytheweb: I think that is a good idea, however I'm not sure if this would work correctly with translations. for instance if id is /request/new will this work without tricky workarounds when url is /pt/request/new?

web-client/src/modules/request/index.ts
puzzledbytheweb: Have you checked this out? https://ant.design/components/icon/

web-client/src/pages/routes/DashboardRoute.tsx
puzzledbytheweb: I like this idea. Having everything controlled by a JSON or a JS object, without needing to change the code to perform modifications or to rely on tricky automations

web-client/src/modules/request/index.ts
puzzledbytheweb: I do think that prop drilling in these cases make the data flow clearer. I like using Context for more global things such as themes

web-client/src/types/menu-item.d.ts
puzzledbytheweb: I'm no expert but I think having it declared there makes it easier to find and modify. If we end up using this in another place we can always move it to a more global file

web-client/src/components/DashboardLayout/DashboardLayout.tsx
puzzledbytheweb: Since this is an interface declaration I'd say it should be in a more global file, so you don't have to repeat it. It should then be available as a type in other files

web-client/src/components/SideDrawer/SideDrawer.tsx
puzzledbytheweb: Yeah. Can you take care of that?

btuerker: Done by 72464e5

web-client/src/modules/request/index.ts
btuerker: Yes, the icons here I passed are exactly the antd icons.

If I use like below:
```javascript
{
  id: uniqueId(),
  icon: FileAddOutlined,
}
```
Then, I have to createElement from that with the same in render method:
```javasript
<Menu.Item key={item.id} {...other}>
   {item.icon && React.createElement(item.icon)}
   {item.title}
</Menu.Item>
```

Also, I've just checked from react-router-dom source code and saw they handled it with the similar way.

```javascript
<Route path={''} component: SomeComponent />
// correspondent block from source code
component
  ? React.createElement(component, props)
  : render
```

web-client/src/components/DashboardLayout/DashboardLayout.tsx
btuerker: Refactored in a82010c

web-client/src/modules/request/index.ts
puzzledbytheweb: Ok, thanks for explaining :)

web-client/src/types/menu-item.d.ts
btuerker: You're right. If we need to use these declarations somewhere else then we can consider making them global. It's really better now, thanks!

Done by a126122

ZiyangGao1028:
Hi Burhan,

Thanks for taking the initiative! The Dashboard looks amazing! Have you passed all the checks recently? From the merge review I saw that there is one check left to pass before it's ready to merge.

Best,
Ziyang

btuerker:
> Hi Burhan,
> 
> Thanks for taking the initiative! The Dashboard looks amazing! Have you passed all the checks recently? From the merge review I saw that there is one check left to pass before it's ready to merge.
> 
> Best,
> Ziyang

The failing check is just the pull request labeler. Not a big deal. But before a merge, I'm waiting for a review from @ashwinkjoseph as well as there're some improvements required as I commented.

web-client/src/pages/routes/DashboardRoute.tsx
sabind: Sounds like a good idea.

web-client/src/pages/routes/DashboardRoute.tsx
sabind: I’d argue the code is easily understandable as is and the compiler will catch issues. I don’t see a need to abstract this away.

web-client/src/types/menu-item.d.ts
sabind: Just because some this is reused only once doesn’t mean we shouldn’t place it more globally if we expect to want/need it in the future. If things are hidden then they won’t be reused in the future.

web-client/src/pages/MasterPage.tsx
sabind: If someone copies they dashboard code they won’t see it working correctly unless they also change this code. 

> As we discussed with @ashwinkjoseph this is not required at all because most of our pages will use the same layout except login steps. But I couldn't figure out how to merge DashboardRoute and ProtectedRoute at this moment. That's why I used this for just experimentation. Maybe the DashboardRoute is unnecessary, then we can just use only the ProtectedRoute.

@ashwinkjoseph do have any insights on how to pull this off?

web-client/src/modules/request/index.ts
sabind: We don’t have to match by string directly. We could use a regex to check the location vs an exact match.

web-client/src/components/SideNavMenu/SideNavMenu.tsx
sabind: What’s the error you’re trying to fix by using `ts-ignore`?

web-client/src/modules/request/index.ts
ashwinkjoseph: this is fine and I feel is required. so let it be

web-client/src/modules/request/index.ts
ashwinkjoseph: I have fixed this issue in a commit on this branch.
I don't think it's a good idea to put this logic in the modules.
It's better to put this logic into the dashboard layout component and thus abstract it away.
The use of `React.createElement` is fine, but it should be in the dashboard layout component itself.
using `React.createElement(x)` is the same as using `<x />` in JSX and in this particular case since it's an object and we aren't using React to render this, you have to use `React.createElement`, however, in the method I used in the latest commit, this is no longer required and am using JSX.

web-client/src/pages/MasterPage.tsx
ashwinkjoseph: yea, I don't think that it should be done in this manner, I am refactoring this a little bit.
the approach is right, but as we discussed, to get this right we have to have changes in the `ProtectedRoute` component as well.
The two should be independent and modules should be able to specify independently whether it should be protected or not irrespective of whether it's having this layout.

ashwinkjoseph: I don't fully understand what you meant @sabind but I will refactor this a little bit and can you review it after that and let me know if it  solves the issue you mentioned?

web-client/src/components/SideNavMenu/SideNavMenu.tsx
ashwinkjoseph: @sabind the issue here is that `react-app-location` doesn't have a type definition and so when he specified the type of item.location as `Location`, typescript considers it to be `window.location`.
I fixed this issue by using { path: string } as the type for item.location.

web-client/src/pages/routes/DashboardRoute.tsx
ashwinkjoseph: yea, @btuerker this is fine, the only part I asked to be in configuration was actually the links and the menu items themselves.
You have done that correctly.
You don't have to generalize this part of the menu as it's part of the design and will remain the same.
although if you are generalizing this part as JSON, it should have been a format similar to the one used for the menu items themselves.

ashwinkjoseph: However, there are a few issues with this again.
1. As per the designs, the center button should be for a new request and not for the map. You have currently put the link to the map there.
2. I think @telmodias wanted the new Request page to be modal and so it should change a state variable which should show the new request modal.
3. The notifications shouldn't be a page, it should be just like the side menu. The reason being, when people visit this website from a desktop browser, they would prefer it to be like a hovering menu that shows them the notifications rather than being redirected to a new page.

@puzzledbytheweb @telmodias could you please confirm these?

ashwinkjoseph: @sabind what do you think about my opinion on the profile details, sign up buttons, etc being static? https://github.com/reach4help/reach4help/pull/481#discussion_r419085948
or do you think it should be a JSON, I think only the menu items should be a JSON

web-client/src/types/menu-item.d.ts
ashwinkjoseph: yup, I agree with this

web-client/src/modules/request/index.ts
ashwinkjoseph: hey, @btuerker , how do you plan to implement the default selected item?
if I remember correctly, we are implementing the` menu item list` and the `menu items` with our own `recursive logic` right?
so I don't see how defining default is going to be a problem here

web-client/src/pages/MasterPage.tsx
ashwinkjoseph: @sabind can you check this part of the code in https://github.com/reach4help/reach4help/pull/481/commits/3386f42c4314c4fd96aed293445cfc25f11c3370 
would this be fine or do you think we need to make changes to it?
I was trying to do it as children, but then had issues with sending the children to the <Route> as a component prop as the type they were expecting was conflicting with the type for react children

sabind: Yea I was thinking in other libs/frameworks that these are done with child views and stacked layouts. If that’s not the case then that’s fine, but this does seem weird with a check about which layout should be rendered.

web-client/src/pages/routes/DashboardRoute.tsx
sabind: Honestly maybe for now it’s best to keep explicit as it is already defined that way. I’m hesitant to keep increasing the file count for the sake of abstraction and it takes more effort to figure out where the code you need to change is buried (when you have a lot of code in between).

I like making it all explicit, but I also don’t want to impose. Just leaving my 2 cents.

web-client/src/pages/MasterPage.tsx
ashwinkjoseph: Yea, I was trying to do the same but was running into `type` issues as the `type definition` of `children` is not compatible with what the `Route` component expects.
But I have the next best solution for this. I will push this change now

web-client/src/pages/routes/DashboardRoute.tsx
ashwinkjoseph: @sabind I am not sure what you meant by keeping it explicit.

In the case of the these items in the sidebar, it's static but the content is also pretty much fixed, but the designs themselves change based on the device and the type of user but not the modules.
So I think by converting these into JSON configs, we are trying to solve a problem that doesn't exist.

I also feel that if you accept these as JSON configs, the content should be customizable by each module, while the designs remain the same across devices which is the opposite of this particular case.

The reason why I suggested a JSON configuration method for the Menu Items was because, the content in it (the menu items) could change from module to module and I wanted to keep one component for the Layout completely and not have it in every single module, so I needed to define it as JSON.
And this was easy to implement as there was a fixed structure to it and could be implemented using the recursive method we discussed a few weeks back.
 
apart from that there, are the changes I requested in  https://github.com/reach4help/reach4help/pull/481#discussion_r419086837

puzzledbytheweb:
closes https://github.com/reach4help/reach4help/issues/385

PR #483, opened 2020-04-28
Map: Implement new design and responsiveness

s0:
## Description

(opening draft now so that we can deploy this branch and people can play around with it while it's still in-development)

Implement the new design and responsive functionality for the map.

## Release Checklist

## Additional Notes

Closes: #75


puzzledbytheweb:
Shit. I clicked "Ready for Review" because I read "Ready for Preview" and hoped for some reason that it would direct me to the preview of the map -.- 

Can you undo that? @s0

s0:
> Shit. I clicked "Ready for Review" because I read "Ready for Preview" and hoped for some reason that it would direct me to the preview of the map -.- 
> 
> Can you undo that? @s0  

Haha, no there's no way to re-draft it unfortunately, that's okay we'll just keep it open! :-D

s0:
Okay, i think enough of the design is implemented now that we can get this merged and deployed. the experience is already significantly better on mobile, and the only parts of the design that aren't done are the results panel, and the mobile filters.

map/src/styling/index.ts
```diff
@@ -14,8 +14,23 @@ const {
 export { css, createGlobalStyle, keyframes, ThemeProvider, THEME };
 export default styled;
 
+const LARGE_BREAKPOINT = 1000;
```
puzzledbytheweb, line 17: We have a file here with all the breakpoints, maybe we can derive a way to use this instead? - https://github.com/reach4help/reach4help/blob/master/web-client/src/constants/mediaQueries.ts

s0: Ah, sharing between `web-client` and `map` requires introducing a new package with a compile step (or using symlinks). I think it's something we want to do, but probably shoudln't block this going in based on sharing of these constants (in particular given that the breakpoints are different right now).

puzzledbytheweb: Yeah, I agree. I'll open up a new issue though so we are aware of this later

PR #484, opened 2020-04-29
[WIP] Disable hiding address bar in mobile

s0:
## Description

Not tested yet, opening PR now to make testing on my mobile easier :)

To go in after #483

## Motivation

When the size of the page is based on the potential max size (100%) of the viewport with the address bar hidden, then users may accidentally scroll the whole page up along with the address bar, which is not desirable. We should base the size on the page on the available space when the address bar is visible.


s0:
pushed commits to #483

Issue #386, opened 2020-04-11
[New Request] - Add "request created" page


jpmarques66:
Screen to display after PIN creates a request.

**Describe the solution you'd like**

Tasks:
- [x] Create component
- [x] Integrate with New-request page #162 

**Figma**: https://www.figma.com/file/OiU7Tl4k1YFfqZQXxLaRrl/Reach4Help?node-id=756%3A145

![image](https://user-images.githubusercontent.com/52417176/79050406-b0c9c680-7c21-11ea-9546-3636817ffc14.png)

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/386)

*This information is updated automatically. To modify it, please use ZenHub.*

**Belonging to Epics:**

* **[EPIC]** [#330 - View: PIN: Request something new](https://github.com/reach4help/reach4help/issues/330)

puzzledbytheweb:
@ashwinkjoseph Is this done? If not do you think we can make these issues (good first issue) really visible for new people on Slack?

ashwinkjoseph:
no, This issue hasn't been done yet.

Issue #348, opened 2020-04-08
Implement View: PIN: Request created screen (please wait)


s0:
* [ ] Implement visuals with dummy data
* [ ] Integrate with redux and firebase

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/348)

*This information is updated automatically. To modify it, please use ZenHub.*

**Belonging to Epics:**

* **[EPIC]** [#331 - View: PIN: Request created screen (please wait)](https://github.com/reach4help/reach4help/issues/331)

**Blocked By:**

* ~~[#348 - implement component: layout / header + side drawer](https://github.com/reach4help/reach4help/issues/324)~~
* ~~[#348 - Design view: PIN: Request created (please wait)](https://github.com/reach4help/reach4help/issues/347)~~

Issue #332, opened 2020-04-08
View: PIN: My Requests


s0:
--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/332)

*This information is updated automatically. To modify it, please use ZenHub.*

**Belonging to Epics:**

* **[EPIC]** [#329 - Views: PIN Flow](https://github.com/reach4help/reach4help/issues/329)

**Children:**

* [ ] [#102 - Implement View: PIN: My Requests](https://github.com/reach4help/reach4help/issues/102)
* [x] ~~[#353 - Design View: PIN: My Requests](https://github.com/reach4help/reach4help/issues/353)~~

PR #486, opened 2020-05-01
New request success page

puzzledbytheweb:
## Description
This is the page to where a PIN is redirected after they successfully submit a request.

So far it's only styled for mobile
## Release Checklist

## Security Checklist

## Additional Notes
image

closes https://github.com/reach4help/reach4help/issues/386
closes https://github.com/reach4help/reach4help/issues/332
related https://github.com/reach4help/reach4help/issues/348


puzzledbytheweb:
> Looks awesome!!!!
> 
> Only one open question since I’m not familiar with the full web client app. Is it common to push to the browser history to change the route?

Hey @sabind ! We're using React Router 5 and in this last update they introduced some react hooks, useHistory is one of these. From [here](https://dev.to/finallynero/hooks-introduced-in-react-router-v5-1-7g8):

> The useHistory gives us access to the history object which helps us programmatically navigate or change routes.

So it is one of the ways we could do this. They also have a <Redirect /> component but I think this way is easier and clearer. It is also used in other places in our codebase :)

sabind:
Sounds good :)

puzzledbytheweb:
closes https://github.com/reach4help/reach4help/issues/348

PR #502, opened 2020-05-05
Backend: introduce a REST API that allows people to consume our map data

### LINKED ISSUE #487: Map: provide an API that allows people to consume our data.
(and probably an NPM package too)

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/487)

s0:
## Description

This is based on #501 (so should get merged after)

Introduces a new firebase "host" at api.reach4help.org (see: https://api.reach4help.org/edge/map/data) for creating general API endpoints for interacting with other projects.

## Motivation

As part of our collaboration with mutualaid.wiki, they need an endpoint to be able to consume our data. Once the mutual aid data has been migrated over to mutualaid.wiki, and we instruct users to enter new group information there instead of using our form, then it's unlikely there will be a continued need for mutualaid.wiki to ue this API, but it will remain useful for others!

(Note: we still need to transfer the hard-coded data to our database, hence the low result count)

## Release Checklist

## Security Checklist

cc @reach4help/security 

## Additional Notes

Closes #487

The majority of our data still needs to be migrated from JSON to our firestore database before it becomes available through this API.

cc @tapjay @Pingid

### DIFF
--- a/functions/src/https/api/map/index.ts
+++ b/functions/src/https/api/map/index.ts
@@ -48,5 +48,6 @@ export const data = functions.https.onRequest(async (_req, res) => {
     }
   });
   res.setHeader('Content-Type', 'application/json');
+  res.setHeader('Cache-Control', 'public, max-age=300, s-maxage=600');
   res.status(200).send(JSON.stringify(result));
 });
--- a/functions/src/https/api/map/index.ts
+++ b/functions/src/https/api/map/index.ts
@@ -1,0 +1,47 @@
+import * as functions from 'firebase-functions';
+
+import { db } from '../../../app';
+
+import { IMarker, MARKER_COLLECTION_ID } from '../../../models/markers';
+
+/* eslint-disable max-len */
+const LICENSES = {
+  reach4help:
+    'This data by Reach4Help (https://reach4help.org/) is licensed under a Creative Commons Attribution-NonCommercial-ShareAlike 4.0 International License (http://creativecommons.org/licenses/by-nc-sa/4.0/)',
+  // TODO: uncomment when license is confirmed with mutualaid.wiki
+  // 'mutualaid.wiki': `This data by Mutual Aid Wiki (https://mutualaid.wiki/) is licensed under a Creative Commons Attribution-NonCommercial-ShareAlike 4.0 International License (http://creativecommons.org/licenses/by-nc-sa/4.0/)`
+};
+/* eslint-enable max-len */
+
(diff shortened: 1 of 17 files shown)


functions/src/https/api/map/index.ts
```diff
@@ -0,0 +1,52 @@
...
+    // TODO: add other sources (with appropriate licenses)
+    if (!docData.source) {
+      result.data.push({
+        ...docData,
+        id: doc.id,
+        license: 'reach4help',
+      });
+    }
+  });
+  res.setHeader('Content-Type', 'application/json');
```
sabind, line 50:
```suggestion
  res.setHeader('Content-Type', 'application/json');
  res.setHeader('Cache-Control', 'public, max-age=300, s-maxage=600');
```

`max-age` is for browsers `s-maxage` is for the CDN.

sabind: I suggest longer than `10` minutes, maybe a few hours at least.

sabind: I would also suggest paginated replies, but that won't affect us so much as end user experience: https://firebase.google.com/docs/firestore/query-data/query-cursors

I'd recommend using the last returned document as the pagination token and simply `startAt(lastDocumentFromPreviousPage)` since document IDs are indexed and it's quite easy to combine that with filters.

sabind:
Since we only have one endpoint we don't need to consider this now, but in the future this might be something to think about. https://nextjs.org/docs/api-routes/introduction

Again, this is an improvement for down the road and shouldn't impact this PR.

functions/src/https/api/map/index.ts
s0: 10 minutes is probably fine for now tbh, we want the data to be fresh :)

Pagination could be good, but given that this data is primarily going to be consumed by other projects in a data-sciency manner (rather than being used directly in frontend projects (at least I hope)), i think this is fine for now :)

s0: 10 minutes is probably fine for now tbh, we want the data to be fresh :)

Pagination could be good, but given that this data is primarily going to be consumed by other projects in a data-sciency manner (rather than being used directly in frontend projects (at least I hope)), i think this is fine for now :)

s0:
@sabind accepted your caching suggestion :)

s0:
Going to merge this in now as i'm working on other stuff that depends on this.

PR #514, opened 2020-05-07
Map: Hide some on the embedded map

### LINKED ISSUE #368: Allow for embedding of map in other pages
**Is your feature request related to a problem? Please describe.**
Vytality Health would like to embed our map on their [Co-Caring Dashboard](https://co-caring-dashboard.netlify.com/). As it currently stands, embedding the map in an iframe would include a bunch of extra clutter which will get in the way.

MutualAid.world have also expressed interest in doing this.

**Describe the solution you'd like**
We should allow the map to display a more "streamlined" version when embedded in an iframe, you could use this technique: https://stackoverflow.com/questions/326069/how-to-identify-if-a-webpage-is-being-loaded-inside-an-iframe-or-directly-into-t

* Completely hide the header + navigation
* Display a new (minimal) header with a link to `map.reach4help.org` that opens in a new window (or `_top`) and the same text as the title of the page.

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/368)

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/368)

btuerker:
## Description
1. Change the header to just show title and a link refers to the original page with less height(it doesn't look much good with the original header height). 

2. Skip about page when the page is in a frame.

Here's a simple example case, maybe we can use the same in documentation as a snippet: 
```html
<!-- This is strictly required for responsiveness, if not the iframe page will not be responsible on mobile screens -->
<meta name="viewport" content="width=device-width,initial-scale=1,user-scalable=no"/>
<style>
  .frame-wrapper {
    position: relative;
    padding-top: 100%; /* 1:1 aspect ratio */
  }

  .frame-wrapper iframe {
    position: absolute;
    top: 0;
    left: 0;
    width: 100%;
    height: 100%;
  }
</style>

<div class="frame-wrapper">
<!-- Without the allow geolocation rule like below, My Location button will not even ask browser for geolocation permission -->
  <iframe src="https://map.reach4help.org" allow="geolocation"></iframe>
</div>
```

## Release Checklist

## Security Checklist

## Additional Notes

Screen Shot 2020-05-07 at 11 34 26 PM

closes #368


map/src/util/index.ts
```diff
@@ -15,3 +15,11 @@ export const isReferrerFromBaseSite = (): boolean => {
   }
   return referrerUrl.origin === 'https://reach4help.org';
 };
+
+export const isInFrame = (): boolean => {
+  try {
+    return window.self !== window.top;
+  } catch (_) {
+    return false;
```
s0, line 23:
```suggestion
    return true;
```

This should be `true` not `false`, as it will fail when `top` and `self` do not have the same origin, and hence are not equal.

map/src/i18n/langs/tr.json
```diff
@@ -1,5 +1,6 @@
 {
   "title": "COVID-19 Karşılıklı Yardımlaşma Haritası",
+  "inFrameTitle": "Bu haritayı {map} üzerinde göster",
```
s0, line 3:
```suggestion
```

adding translations directly in these json files will be removed by crowdin, we need to make sure that translations are only added using that UI (i.e. after the new english strings are added to `master`)

map/src/i18n/langs/en.json
```diff
@@ -1,5 +1,6 @@
 {
   "title": "COVID-19 Mutual Aid Map",
+  "inFrameTitle": "View this map on {map}",
```
s0, line 3: In most cases, there won't be the context of Reach4Help when embedded, so people won't necessarily understand that this link will allow them to find more information about the map, or add information.

I suggest we do this instead:

```suggestion
  "inFrameLink": "About this map / add information",
```

map/src/components/header.tsx
```diff
@@ -77,7 +78,37 @@ class Header extends React.Component<Props, State> {
...
+                      <a
+                        key={key}
+                        href="https://map.reach4help.org"
+                        target="_blank"
+                        rel="noopener noreferrer"
+                      >
+                        Reach4Help.org
+                      </a>
+                    ),
+                  })}
```
s0, line 104: Let's make this whole test a link using the new text I suggested.

map/src/components/header.tsx
```diff
@@ -77,7 +78,37 @@ class Header extends React.Component<Props, State> {
...
+    const { className, page, setPage, inFrame } = this.props;
+
+    if (inFrame) {
+      return (
+        <AppContext.Consumer>
+          {({ lang }) => (
+            <header className={className}>
+              <div
+                className="top"
+                style={{ height: '36px', justifyContent: 'center' }}
```
s0, line 90: also not fond of this inline styling, i'll fix this too :)

PR #515, opened 2020-05-07
Map: incorporate mutualaid.wiki data in our map

s0:
## Description

This is quite a heafty PR. The overall effect of which is to include the 3000+ groups from https://mutualaid.wiki in our map's data-set, and update the UI with attribution, and direct people to add mutual aid groups on that site instead of ours from the Add information window.

A few changes needed to be made to accommodate the vast increase in datapoints on the map. The main changes in this PR are:

* Introduction of a new cloud function to sync with mutualaid.wiki (needs to be converted to cron)
* Introduction of a new cloud function to cache the contents of the markers collection as a JSON file (in firebase storage) that can be downloaded by clients instead of directly reading the 3000+ entries from firebase (which is a lot slower, and eats into our usage)
* Updating the Map to download the marker data from the firebase storage instead of firestore.
* Updated the API to spit out the lat/longs without the underscores (these were the hidden properties of the GeoPoint class in firebase).
* Updated the map to handle points with no radius
* Updated the results in the map to only display a limited number at a time, and added a button at the bototm to show more results
* Added attribution information to the details view of all markers that come from this source (with a link to mutualaid.wiki)
* Added a disclaimer message to the details pane of all results.
* Added instructions to the add information window when selecting mutual aid organizations to submit to mutualaid.wiki instead
* Added information in the about page about our data, licenses and API endpoint.

## Release Checklist

## Security Checklist

### DIFF
--- a/firebase/storage.rules
+++ b/firebase/storage.rules
@@ -10,6 +10,12 @@ function isUser(id) {
 
 service firebase.storage {
   match /b/{bucket}/o {
+
+    match /map-markers-public/{allPaths=**} {
+      allow read: if true;
+      allow write: if false;
+    }
+
     // No one can do anything with the storage layer
     match /{allPaths=**} {
       allow read, write: if false;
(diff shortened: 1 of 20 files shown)


s0:
I'm going to merge this now just to get it deployed ASAP, but it could still do with some review @sabind

s0:
Also security rules are updated here @reach4help/security

PR #519, opened 2020-05-08
Map: schedule the recently created map functions

s0:
## Description

Until we do this, we'll need to run them manually.

## Release Checklist

## Security Checklist


sabind:
I noticed there's some general processing changes we can make to improve the number of reads + writes used, but it would require additional changes on the mutual aid wiki side potentially so for now this looks really good.

PR #542, opened 2020-05-13
Add roles collection and logic to link them to claims

sabind:
## Description

When users have records added to the roles collection a function fires to update the user's JWT claims in the auth database as well as recording an audit log of the change.

This PR also tidies up some issues with the model objects in the `web-client`

## Motivation

Allow us to manage user permissions.

## Release Checklist

## Security Checklist


functions/src/roles/functions.ts
```diff
@@ -36,14 +36,17 @@ export const onWrite = (change: Change<DocumentSnapshot>, context: EventContext)
...
-  // go over all claims and set them
-  const afterObject = after.toObject() as { [id: string]: string[] };
-  Object.keys(afterObject).forEach(permissionGroup => {
-    (afterObject[permissionGroup]).forEach(permission => {
-      // set the data to true or null (the only two options)
-      newUserClaims[`admin.${permission}`] = true;
+
+  if (after) {
+    // go over all claims and set them
+    const afterObject = after.toObject() as { [id: string]: string[] };
```
s0, line 42: This type you're casting `after.toObject()` to is a little dangerous, and shouldn't be neccesary.

I guess you probably did it due to a type error when trying to do `afterObject[permissionGroup]`, in which case, whitelisting the allowed permission groups as a const would be a way to get around this issue without breaking type safety. :)

functions/src/roles/functions.ts
```diff
@@ -36,14 +36,17 @@ export const onWrite = (change: Change<DocumentSnapshot>, context: EventContext)
...
+    Object.keys(afterObject).forEach(permissionGroup => {
+      (afterObject[permissionGroup]).forEach(permission => {
+        newUserClaims[`${permissionGroup}.${permission}`] = true;
```
s0, line 45: Aha, i see you already fixed the problem here :)

s0:
I'll have a go at implementing my suggested changes for you (especially around removal of the class and various typescript nuances). :)

functions/src/roles/functions.ts
```diff
@@ -0,0 +1,66 @@
...
+    .catch(error => console.error('User Claims Failed: ', error));
+
+  const operations = [
+    db.collection(ROLE_AUDITS_COLLECTION_ID).add({
+      affectedUser: db.collection('users').doc(userId),
+      actingUser: actingUserRef,
+      previousRole: before,
+      currentRole: after,
+      updatedAt: Timestamp.now(),
+    }).catch(error => console.error('Audit Log Failed: ', error)), // Fail silently
```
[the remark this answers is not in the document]
s0: Oops, and indeed there's a mistake as a result here, `affectedUser` -> `affectedUserRef` and `actingUser` -> `actingUserRef`.

s0:
Oooh, we can actually do one better than defining our own type-checking functions, and use https://github.com/gcanti/io-ts to define our types + type-checking functions simultaneously! I'll switch over these classes to that, and we can migrate the other ones later.

s0:
Okay, i've addressed all my comments, please take a look at what I've done and check that you agree.

The biggest change has been the move from `class-validator` to `io-ts`, which I have just discovered and am alreay in love with! :joy:

s0:
Also just tested this on dev, and seems to work well :)

PR #543, opened 2020-05-13
fix: updated ci actions

sabind:
## Description

Updated deploy scripts.

## Release Checklist

## Security Checklist


.firebaserc
```diff
@@ -28,21 +31,27 @@
...
+        ],
+        "site": [
+          "reach4help-staging"
         ]
       }
     },
     "reach4help": {
       "hosting": {
         "web-client": [
-          "reach4help-web-client"
```
sabind, line 37: When the project moved we lost this name to the old `reach4help-xxxx` project :(

s0:
Hmm, this looks good, but i don't know whether this branching + environment mapping makes sense. Notably, i don't think auto deploying to dev is a good idea, isn't that environment supposed to be for us to run and test things while we're developing them, that makes it a bit harder to do if it's automatically deployed to by master right?

And regarding the other branches, why not just have them named `staging` and `production` vs `release` and `prerelease` to simplify it?

sabind:
@s0 I could see using `staging` and `production` branches vs. using the tags and releases features. I don't have the permissions to create the new protected branches, but if you'd like to make them I can update the scripts to leverage them.

s0:
> I don't have the permissions to create the new protected branches, but if you'd like to make them I can update the scripts to leverage them.

That'll be because they already exist :)

Any thoughts on not auto-deploying master to dev too?

sabind:
We can push development -> dev, staging -> staging, prod -> prod and then master is the just final place where we archive the code after successful releases?

s0:
Uhh... I'm still convinced that `dev` should be for manual deployments only, we need an environment where we can test things during development (before PR), and during review (after PR), if not Dev, then we need to create another environment specifically for this purpose.

sabind:
If we keep dev for manual deploys then we won't have anywhere to work on things as a team. If you need ephemeral deploys for testing PRs then users can simply deploy to their own Firestore env and add the URL in their PR. Some deploys to dev may cause data that isn't backwards compatible so deploying to that environment should be regulated until after PRs are approved.

I recommend we go with this flow where we have one branch per env and if we need pull request builds then we can work on that as a precursor to dev (which also follows the branches).

s0:
> If we keep dev for manual deploys then we won't have anywhere to work on things as a team. If you need ephemeral deploys for testing PRs then users can simply deploy to their own Firestore env and add the URL in their PR. Some deploys to dev may cause data that isn't backwards compatible so deploying to that environment should be regulated until after PRs are approved.
> 
> I recommend we go with this flow where we have one branch per env and if we need pull request builds then we can work on that as a precursor to dev (which also follows the branches).

Okay cool that's reasonable. Users making their own projects, or us creating them when needed isn't crazy.

In that case, i'm going to be a pain again and say we should use `master` -> `dev`, not `development` -> `dev`, the development branch's sole purpose is to accept PRs that come from forks, as unfortunately github actions don't run on those PRs, and we need those checks to be required. :)

s0:
I'll revert your PR back to `f975847` and merge :)

sabind:
Ok this makes sense - PRs -> Dev then we merge Dev -> Master to deploy, then we have staging and prod for the remaining envs.

PR #550, opened 2020-05-14
Add actor and action variables to timeline item

btuerker:
## Description
1-) Added two more class variables to the timeline item model that can make front-end implementation easier and in less complexity.

2-) Made offerRef and offerSnapshot optional. I'm going to explain the reasons in below motivation.

Note: I used different naming convention for enum declaration because by common convention, constants are identified by upper case. Is it better to use a lower case definition? 

## Motivation
1-) I put a user reference as `actor` and an enum 'action' because in front-end we'd like to know who is responsible for that timeline-item creation as well as information to get which type of information affected within that timeline-item.

2-) It's not strictly required but in my opinion, it makes the model more consistent. Let's say a pin user created a request and it's now pending offers. We don't need timeline-items at that point(it might be changed later), as we're just showing a list of offers at the timeline screen. But, if we don't record it as a timeline item, in the timeline-items list when it assigned to an offer we always need to put a hardcoded list item for `REQUEST_CREATE` action. Also, a pin user can cancel that request without any offer assigned to it. In addition to that, we can implement all the related notification logic into timeline-item rather than request or offer. It would be better to use this as the only place for request&offer flow notification generation.

## Release Checklist

## Security Checklist


functions/src/models/requests/timeline.ts
```diff
@@ -81,6 +128,17 @@ export class TimelineItem implements ITimelineItem {
     this._requestSnapshot = value;
   }
 
+  @IsObject()
```
sabind, line 131: @IsEnum() is better used here. We can pass in the enum to validate against.

web-client/src/models/requests/timeline.ts
```diff
@@ -81,8 +124,10 @@ export class TimelineItem
   // eslint-disable-next-line class-methods-use-this
   toFirestore(modelObject: TimelineItem): ITimelineItem {
     return {
+      actorSnapshot: modelObject.actorSnapshot,
       offerSnapshot: modelObject.offerSnapshot,
       requestSnapshot: modelObject.requestSnapshot,
+      action: modelObject.action,
       createdAt: modelObject._createdAt,
     };
   }
```
sabind, line 133: The file in the front end should follow the same design as the backend. We should be using `toObject`. The code split before these could be aligned. Until we can extract them, we should keep them aligned (using the `backend` one as the example).

web-client/src/models/requests/timeline.ts
```diff
@@ -16,74 +46,161 @@ export interface ITimelineItem extends firebase.firestore.DocumentData {
...
 
-  // eslint-disable-next-line class-methods-use-this
-  toFirestore(modelObject: TimelineItem): ITimelineItem {
+  toObject(): object {
     return {
-      offerSnapshot: modelObject.offerSnapshot,
-      requestSnapshot: modelObject.requestSnapshot,
-      createdAt: modelObject._createdAt,
+      actorRef: this.actorRef.path,
+      offerRef: this.offerRef ? this.offerRef.path : null,
```
btuerker, line 187: @sabind can you review this?

I locally tested the web-client model by using that model to create a document in the database. It's not working without this change. It gets undefined if we use `this.offerRef?.path`. Without that, I was getting an `unsupported field offerRef` in the toFirebase method, because it returns undefined. Maybe we should refactor Request & User models too. 

I faced the same problem in other models as well. When createdAt is optional `createdAt?.toDate()` causes an error too.

Here's a simple example case: 
```javascript
class TimelineItem {
    public ref?: { path:string } | null = null;
    constructor(ref: { path:string} | null = null) {
        this.ref = ref;
    }
}

const item = new TimelineItem({ path: 'hello'});
console.log(item.ref ? item.ref.path : null); // prints 'hello'
console.log(item.ref?.path); // prints 'hello'

const itemWithoutRef = new TimelineItem();
console.log(itemWithoutRef.ref ? itemWithoutRef.ref.path : null); // prints 'hello'
console.log(itemWithoutRef.ref?.path); // prints undefined

````

File: web-client/src/models/requests/timeline.ts (hunk 2 above)
Comment (btuerker) [reply web-client/src/models/requests/timeline.ts]: Done by 4788542

File: functions/src/models/requests/timeline.ts (hunk 1 above)
Comment (btuerker) [reply functions/src/models/requests/timeline.ts]: Done by 4788542

File: web-client/src/models/requests/timeline.ts (hunk 3 above)
Comment (ashwinkjoseph) [reply web-client/src/models/requests/timeline.ts:187]: yes the conditional setting of the value as `null` is fine.

**Comment by sabind** [comment]
> @sabind I was wondering, if we could store not just the action but also the text to be displayed here in the timeline item itself, wouldn't it be better?

We shouldn't store front-end logic in the backend. I'd proceed with this and store that functionality in the front-end. You can even have it as a function of the timeline item itself that uses a switch statement.

File: functions/src/models/requests/timeline.ts [hunk 4]
```diff
@@ -1,19 +1,35 @@
...
+import { IsEnum, IsNotEmptyObject, IsObject, IsOptional, ValidateNested } from 'class-validator';
 import { firestore } from 'firebase';
 
 import { IOffer, Offer } from '../offers';
 import { IRequest, Request } from './index';
+import { IUser, User } from '../users';
 import Timestamp = firestore.Timestamp;
 import DocumentData = firestore.DocumentData;
 import DocumentReference = firestore.DocumentReference;
 import QueryDocumentSnapshot = firestore.QueryDocumentSnapshot;
```
btuerker, line 11: I think we must import these types from '@google-cloud/firestore'. 

I got a silly issue while I was trying to add another operation to create a timeline at the request create trigger. If we try to get ref from a snapshot, i.e 'snapshot.ref', it's a FirebaseFirestore.DocumentReference not a firebase.firestore.DocumentReference. So, it's mismatching and causes type errors.

Did I miss something? @sabind @ashwinkjoseph

sabind: yea so the models use Firebase and the functions use Firebase Admin so yes we have to make sure we import from the @google-cloud/firestore to be consistent.

btuerker: Also, we have to change the type of operations array. `FirebaseFirestore.DocumentReference` returns `Promise<WriteResult>` not `Promise<void>`

sabind: For that one we simply do a `.then(() => return Promise.resolve())` to throw away the `WriteResult`

PR #551, opened 2020-05-15
feat(backend): introduce bot authentication with custom tokens

s0:
## Description

Allow for the creation of bot accounts by introducing an endpoint
that uses pre-shared-tokens to allow bots to create firebase
tokens for themselves, and add a firestore rule allowing the
chatbot to store config.

## Motivation

The initial use case is for our new chatbot (https://github.com/reach4help/chatbot) to store configuration in firestore, namely the token / credentials used to authenticate with the google calendar API. A message is sent over slack to #bot-admin when authentication is required, with a link to google to facilitate the request.


s0:
For now, I've deployed this on reach4help-dev (and the chatbot is currently using that environment to store its tokens). I plan to move this over to prod when done.

s0:
> How is this API exposed? I don't see changes the config changes to make this available.

Not sure exactly what you mean? The bot can write whatever it wants to the firestore document `/config/chatbot`. As long as there is a token in the function config `bot_auth_tokens` (take a look at the config in `reach4help-dev`).

sabind:
I meant the firebase.json changes to expose the API route to hosting so that it's accessible.

PR #554, opened 2020-05-16
Issue 457: Frontend of CAV Flow View - Requests

Winggo:
## Description
RequestList, Header, and ReqestItem components built for requests module. Clicking on requestItem triggers handleRequest function found in container (logic to be implemented later). If user is 'CAV' and request is open, clicking on requestItem instead opens up options for accepting/rejecting request. Accepting/rejecting buttons logic yet to be implemented.  
## Release Checklist

- [y] I have a plan to verify that these changes are working as expected after deploy
- [y] I have a plan for how to revert, rollback, or disable these changes if things are not working as expected

## Security Checklist

## Additional Notes
requests1
requests2


web-client/src/modules/requests/components/RequestItem/RequestItem.tsx
```diff
@@ -0,0 +1,155 @@
...
+            Cannot Help
+          </StyledButton>
+          <StyledButton
+            style={{
+              background: '#52C41A',
+              color: '#FFFFFF',
+            }}
+            onClick={() => handleRequest(true)}
+          >
+            Help {request.pinUserSnapshot.displayName}
```
ashwinkjoseph, line 126: hey @Winggo could you make this more responsive to the length of the text? something like if the text is a longer, then truncate it and show ... at the end like the `ellipsis` of text-overflow.
currently, this button goes to the next line when the text is longer.
for this, you would have to use the grid structure that `antd` provides and put both of these in one row with two columns and the two buttons in each of the columns. 
I have attached an image to show this problem
image

Winggo:
Hey @ashwinkjoseph, all the issues you have pointed out should be resolved now with my recent commit. Accept/reject request buttons now truncate text with ellipsis with antd grid, layout content styling is consistent, font Roboto imported and now used. Also added some css to requestUserPic to profile pic doesn't stretch.

image

web-client/src/modules/requests/components/RequestItem/RequestItem.tsx
Winggo: Just committed changes to fix these issues

PR #564, opened 2020-05-19
chore(web-client): remove model classes and class-validator

s0:
## Description

With the exception of default values,
the functionality of these classes (i.e. data validation) is not
being used at all in the web-client code, so we can just use the
model interfaces and data objects directly.

For default values, I've introduced the concept of "init"
functions that allow some properties to be made optional,
which means it will now be explicit throughout the
codebase which objects are fully-initialized documents,
and which are partial.

Functionality *should* hopefully remain the same after this change.

## Motivation

* Reduces the amount of code for the models considerably
* Removes functionality that wasn't being used (class validation decorators), and which was likely to be out-of-sync with the interfaces.
* Makes it explicit when we're augmenting a partial document with default values (rather than implicitly relying on class constructors to do this). I.e using the `init*()` functions.
* Reduces the amount of code that has to be migrated to the `model` package, making that migration simpler.

### DIFF
--- a/web-client/src/models/util.ts
+++ b/web-client/src/models/util.ts
@@ -1,0 +1,5 @@
+/**
+ * Make the properties K optional in type T
+ */
+export type MakePartial<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>> &
+  Partial<Pick<T, K>>;
--- a/web-client/package.json
+++ b/web-client/package.json
@@ -20,7 +20,6 @@
     "antd": "^4.0.4",
     "axios": "^0.19.2",
     "babel-plugin-import": "^1.13.0",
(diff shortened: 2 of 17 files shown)


web-client/src/models/util.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * Make the properties K optional in type T
+ */
+export type MakePartial<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>> &
```
puzzledbytheweb, line 4: Here's a snippet of code I'll be keeping close by. This is definitely beyond my Typescript prowess xD

puzzledbytheweb:
> I don't think this is a good idea at the moment.

I agree. We should wait post-MVP.

PR #583, opened 2020-05-23
Add request update actions

btuerker:
## Description
Add an action to update request status and pin&cav ratings.

## Release Checklist

## Security Checklist


web-client/src/ducks/requests/functions.ts
```diff
@@ -75,3 +76,23 @@ export const setUserRequest = async ({
...
+
+  data.updatedAt = firebase.firestore.Timestamp.now();
+  data.pinRating && (data.pinRatedAt = timeNow);
+  data.cavRating && (data.cavRatedAt = timeNow);
+
+  return firestore
+    .collection('requests')
+    .doc(requestId)
+    .update({ ...data });
+};
```
btuerker, line 98: @ashwinkjoseph As I mentioned, I'm not happy with this implementation. However, it works well. I intended to use `setRequest` action but had issues with the request converter I guess.

The issue is simply, we cannot use `toFirebase` with the request stored in redux. Because in the converter `toFirebase` method calls `toObject` and it looks for path variable from a reference, i.e `pinUserRef.path` and it gets undefined because of pinUserRef stored as a string i.e `users/someuid`. 

I'm looking forward to hearing your opinions here. As soon as you mention, I'll work on changes :)

web-client/src/modules/timeline/containers/TimelineViewContainer/TimelineViewContainer.tsx
```diff
@@ -195,25 +229,24 @@ const TimelineViewContainer: React.FC<TimelineViewContainerProps> = ({
...
       Once backend changes for profile snapshot is done, instead of user={mockRequestUser},
       The Top Panel must take the user details from the request itself
   */
+  const isCav =
+    profileState.profile.applicationPreference === ApplicationPreference.cav;
+
   return (
     <>
       <TopPanel request={request} user={mockRequestUser} />;
       <TimelineList items={mockTimelineItems} currentUser={mockPin} />
```
ashwinkjoseph, line 243: @btuerker why are you still mocking these? I thought you were integrating this with the timeline actions as well

web-client/src/ducks/requests/functions.ts
ashwinkjoseph: have you been using the instance of the request class to modify this?
if not that could be the problem

web-client/src/modules/timeline/containers/TimelineViewContainer/TimelineViewContainer.tsx
ashwinkjoseph: since we are a little short on time as this was to be completed today, I am taking care of this integration now.

web-client/src/ducks/requests/functions.ts
btuerker: Fixed 8af5a85, now it's good to go

sabind:
@btuerker - The timeline should be in to the development branch. Lets resolve these conflicts.

Issue #572, opened 2020-05-21
[BUG] SignUp: Validation error message for not checking the "I understand..." checkbox should be clearer


PhilArsHermit:
**Describe the bug**
If the user presses the "Continue" button without checking the "I understand and acknowledge the Terms & Conditions." checkbox, a message will pop simply stating "Missing Data!".  Which doesnt tell the user which fields/checkbox he missed.

**To Reproduce**
1. In the "Please fill out the personal details below." page, enter every fields, but leave the "I understand..." unchecked
2. Press the "Continue" button

**Expected behavior**
The Validation error message should tell the user that he needs to check the "I understand..." checkbox before pressing "Continue"

![Validation_Message_Missing_Data](https://user-images.githubusercontent.com/65713824/82588848-564d4f80-9b69-11ea-9d80-c205b56acbd8.png)

**Desktop (please complete the following information):**
 - OS: W10 pro
 - Chrome
 - Version 81.0.4044.138

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/572)

*This information is updated automatically. To modify it, please use ZenHub.*

ashwinkjoseph:
Hey, I am not able to reproduce this issue for some reason.
![image](https://user-images.githubusercontent.com/13694998/82935670-0af0d200-9fab-11ea-82a8-be1c4129a098.png)
This is what I get when I try to reproduce the same.
I tried using latest version of Chrome for MacOS and the latest version of Firefox for MacOS and couldn't reproduce this issue.
This was without the fix #587 , with the existing codebase on the existing staging env URL at https://reach4help-staging-web-client.web.app

djohal:
> Hey, I am not able to reproduce this issue for some reason.
> ![image](https://user-images.githubusercontent.com/13694998/82935670-0af0d200-9fab-11ea-82a8-be1c4129a098.png)
> This is what I get when I try to reproduce the same.
> I tried using latest version of Chrome for MacOS and the latest version of Firefox for MacOS and couldn't reproduce this issue.
> This was without the fix #587 , with the existing codebase on the existing staging env URL at https://reach4help-staging-web-client.web.app

This works fine when you click on 'Continue' directly. However, try to check and uncheck the t&c box and then click 'Continue' again. you should be able to reproduce this way.

CorporateInvesther:
Issue Fixed on Staging Build 1.0

PR #587, opened 2020-05-25
Dilpreet/terms condition fix

djohal:
## Description
Added validation for T&C on toggle
## Motivation
#572 

## Release Checklist

## Security Checklist

## Additional Notes
bug-572

### DIFF
--- a/web-client/src/modules/personalData/components/PersonalDataForm/PersonalDataForm.tsx
+++ b/web-client/src/modules/personalData/components/PersonalDataForm/PersonalDataForm.tsx
@@ -298,6 +298,7 @@ const PersonalDataForm: React.FC<PersonalDataFormProps> = ({
       address: newAddress,
       sendNotificatoins: allowSendNotifications ? new Date() : null,
     };
+    console.log(termsAndPrivacyAccepted);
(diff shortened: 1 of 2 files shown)


sabind:
Looks good to me. Will wait for @ashwinkjoseph for the final review.

ashwinkjoseph:
@djohal This covers an issue with the default `required:true` rule of `antd`.
However, this isn't the only place we use this rule at. could you check other places that we use this rule and see if the issue exists there too? 
If so could you look into fixing those as well? we can merge all of that together.
If this isn't a problem anywhere else, we can merge this right now

djohal:
> @djohal This covers an issue with the default `required:true` rule of `antd`.
> However, this isn't the only place we use this rule at. could you check other places that we use this rule and see if the issue exists there too?
> If so could you look into fixing those as well? we can merge all of that together.
> If this isn't a problem anywhere else, we can merge this right now

I couldn't find any other similar issue. However, I did find couple of issues with phone number validation, it wasn't taking in the regex validation and was only checking for minimum length. I updated it to show an error if it doesn't match the format put in the placeholder along with the minimum length check.

**Before:**
image

**After:**
image

PR #596, opened 2020-05-27
Feat/notifications frontend integration

Winggo:
## Description
Completed notifications feature, added `seenAt` and `requestSnapshot` attributes to offers model in web-client/ and functions/, updated existing `setOffer` actions' payload to include `requestSnapshot` attribute

Notifications is powered by the offers collection, sorted by `seenAt` date and displays messsage involving requestAuthor, requestTitle, offerAuthor, offerStatus
## Motivation
Users can now view history and live updates of requests followed.
## Testing Guidelines
Due to the new `requestSnapshot` and `seenAt` attributes being added to the Offer class, please fill in those fields when testing. Those fields are marked optional so that existing datasets remain operational. If `seenAt` for an offer is null, that offer will not be displayed in notifications due to the offer having no activity. 
## Release Checklist

- [y] I have a plan to verify that these changes are working as expected after deploy
- [y] I have a plan for how to revert, rollback, or disable these changes if things are not working as expected

## Security Checklist

## Additional Notes
image
image


functions/src/models/offers/index.ts
```diff
@@ -19,9 +20,11 @@ export interface IOffer extends DocumentData {
   pinUserRef: DocumentReference<DocumentData>;
   requestRef: DocumentReference<DocumentData>;
   cavUserSnapshot: IUser;
+  requestSnapshot: IRequest | null;
   message: string;
   status: OfferStatus;
   createdAt?: Timestamp;
+  seenAt?: Timestamp;
```
ashwinkjoseph, line 27: This should also be allowed to be `null`. firebase doesn't play well with `undefined`

functions/src/models/offers/index.ts
```diff
@@ -30,17 +33,21 @@ export class Offer implements IOffer {
     pinUserRef: DocumentReference<DocumentData>,
     requestRef: DocumentReference<DocumentData>,
     cavUserSnapshot: User,
+    requestSnapshot: Request | null,
     message: string,
     status: OfferStatus,
     createdAt = Timestamp.now(),
+    seenAt = Timestamp.now(),
```
ashwinkjoseph, line 40: You shouldn't be assigning the default value of `seenAt` as now, because then every offer created would have the value of `seenAt` to be the moment it's created.
instead it should be set to `null` in the constructor

web-client/src/ducks/offers/actions.ts
```diff
@@ -56,10 +58,14 @@ export const setOffer = (payload: Offer | IOffer, offerId?: string) => (
       firebase: offerId ? setUserOffer : createUserOffer,
     });
   } else {
+    const offerPayload = {
```
ashwinkjoseph, line 61: the else condition here means that payload is already an instance of Offer.
So, you must set the value of `seenAt` as `payload.seenAt = firestore.Timestamp.now()` and shouldn't be creating a new object with it.
even if you are creating a new object (which you shouldn't), you should be creating it as: 
`Offer.factory({ ...payload.toObject(), seenAt: firestore.Timestamp.now() })`

web-client/src/models/offers/index.ts
```diff
@@ -28,9 +30,11 @@ export interface IOffer extends firebase.firestore.DocumentData {
     firebase.firestore.DocumentData
   >;
   cavUserSnapshot: IUser;
+  requestSnapshot: IRequest | null;
   message: string;
   status: OfferStatus;
   createdAt?: firebase.firestore.Timestamp;
+  seenAt?: firebase.firestore.Timestamp;
```
ashwinkjoseph, line 37: same as I mentioned in model at functions

web-client/src/models/offers/index.ts
```diff
@@ -45,17 +49,21 @@ export class Offer implements IOffer {
       firebase.firestore.DocumentData
     >,
     cavUserSnapshot: User,
+    requestSnapshot: Request | null,
     message: string,
     status: OfferStatus,
     createdAt = firestore.Timestamp.now(),
+    seenAt = firestore.Timestamp.now(),
```
ashwinkjoseph, line 56: same as I mentioned in the models in the functions

functions/src/models/offers/index.ts
sabind: +1

functions/src/models/offers/index.ts
sabind: +1

web-client/src/ducks/offers/actions.ts
sabind: +1

sabind:
Overall the PR is good. We just need to make sure we're using `null` or `Timestamp` we figured this out earlier that `undefined` would give us some complications so we want to restrict it so comparisons are easy and accurate.

web-client/src/ducks/offers/actions.ts
Winggo: Got it, got this fixed now

functions/src/models/offers/index.ts
Winggo: Gotcha

functions/src/models/offers/index.ts
Winggo: Fixed, I've set it as Timestamp.now() as default value and my reasoning for it was because essentially when an offer is created or updated, an activity is happening at that current moment to generate a notification

But setting the seenAt time in the setOffer action does the same thing and makes more sense
`if (!(payload instanceof Offer)) {
    const offerPayload = Offer.factory({
      ...payload,
      seenAt: firestore.Timestamp.now(),
    });
    dispatch({
      type: SET,
      payload: {
        offerPayload,
        offerId,
      },
      firebase: offerId ? setUserOffer : createUserOffer,
    });
 } else {
    // eslint-disable-next-line no-param-reassign
    payload.seenAt = firestore.Timestamp.now();
    dispatch({
      type: SET,
      payload: {
        payload,
        offerId,
      },
      firebase: offerId ? setUserOffer : createUserOffer,
    });
}
`

web-client/src/models/offers/index.ts
Winggo: Fixed

web-client/src/models/offers/index.ts
Winggo: fixed

PR #598, opened 2020-05-28
fixing getMyLocation.  pans to current center

mcverter:
current GetMyLocation UI also needs to PAN to the location after loading it.


web-client/src/components/WebClientMap/WebClientMap.tsx
```diff
@@ -52,6 +52,11 @@ const WebClientMap: React.FC<MapProps> = ({
   const initGoogleMapServices = ({ map, maps }) => {
     if (map && maps) {
       googleMap || setGoogleMap(map);
+      /*
+      map.controls[google.maps.ControlPosition.RIGHT_CENTER].push(
+        <div style={{ zIndex: 1000 }}>hello world</div>,
+      );
+      */
```
sabind, line 59: I'm guessing this is some accidental left over code?

mcverter: removed

PR #611, opened 2020-06-01
new cloud function along with integration in frontend, minor fixes, firebase rules

ashwinkjoseph:
## Description
The PR completes the new cloud function which will make the code more cleaner and remove the dependency on having complicated/unsafe firebase rules and finishes the integration of this cloud function into the frontend.
This also features abstract classes that can give location and timelines of requests and offers directly in one request
## Motivation
The statuses of the requests being stored as data are different from the screens that we have. A request with the same value for its status can appear in different screens based on factors such as offers made and who is making the request.

This logic is now separated into a cloud function so that the frontend can query based on its abstract status and backend hides the complexity.

This also prevents the frontend from having to query and obtain data that it shouldn't be allowed to (for example, all the timeline events for a request when the CAVs offer was rejected, or the locations of the CAVs when viewing the offers)

This also gives us greater freedom to change the logic in the backend later while maintaining the same interface.

This also helps us maintain and fix bugs in the frontend with more ease than earlier with all the complicated business logic being involved.
## Release Checklist

## Security Checklist

### DIFF
--- a/firebase/firestore.indexes.json
+++ b/firebase/firestore.indexes.json
@@ -1,4 +1,12 @@
 {
-  "indexes": [],
+  "indexes": [
+    {
+      "collectionId": "requests",
+      "fields": [
+        { "fieldPath": "status", "mode": "ASCENDING" },
+        { "fieldPath": "latLng", "mode": "ASCENDING" }
+      ]
+    }
+  ],
   "fieldOverrides": []
 }
--- a/functions/src/models/offers/index.ts
+++ b/functions/src/models/offers/index.ts
@@ -162,5 +162,5 @@ export const OfferFirestoreConverter: FirestoreDataConverter<Offer> = {
     message: modelObject.message,
     status: modelObject.status,
     createdAt: modelObject.createdAt.toDate(),
-  })
+  }),
 };
--- a/functions/src/models/offers/index.ts
+++ b/functions/src/models/offers/index.ts
@@ -189,9 +189,7 @@ export class Offer implements IOffer {
       pinUserRef: this.pinUserRef,
       requestRef: this.requestRef,
       cavUserSnapshot: this.cavUserSnapshot.toObject(),
-      requestSnapshot: this.requestSnapshot
-        ? this.requestSnapshot.toObject()
-        : null,
+      requestSnapshot: this.requestSnapshot ? this.requestSnapshot.toObject() : null,
       message: this.message,
       status: this.status,
       createdAt: this.createdAt,
--- a/functions/src/models/offers/index.ts
+++ b/functions/src/models/offers/index.ts
@@ -1,5 +1,6 @@
+import { FirestoreDataConverter } from '@google-cloud/firestore';
 import { IsEnum, IsNotEmpty, IsObject, IsString, ValidateNested } from 'class-validator';
-import { firestore } from 'firebase';
+import { firestore } from 'firebase-admin';
 
 import { IUser, User } from '../users';
 import Timestamp = firestore.Timestamp;
@@ -124,7 +125,7 @@ export class Offer implements IOffer {
     this._createdAt = value;
   }
 
-  static factory = (data: IOffer): Offer => {
+  public static factory(data: IOffer): Offer {
     return new Offer(
       data.cavUserRef,
       data.pinUserRef,
@@ -134,7 +135,7 @@ export class Offer implements IOffer {
       data.status,
       data.createdAt,
     );
-  };
+  }
 
   toObject(): object {
     return {
@@ -149,7 +150,7 @@ export class Offer implements IOffer {
   }
 }
 
-export const OfferFirestoreConverter: firebase.firestore.FirestoreDataConverter<Offer> = {
+export const OfferFirestoreConverter: FirestoreDataConverter<Offer> = {
   fromFirestore: (data: QueryDocumentSnapshot<IOffer>): Offer => {
     return Offer.factory(data.data());
   },
--- a/functions/src/models/offers/index.ts
+++ b/functions/src/models/offers/index.ts
@@ -154,5 +154,13 @@ export const OfferFirestoreConverter: FirestoreDataConverter<Offer> = {
   fromFirestore: (data: QueryDocumentSnapshot<IOffer>): Offer => {
     return Offer.factory(data.data());
   },
-  toFirestore: (modelObject: Offer): DocumentData => modelObject.toObject(),
+  toFirestore: (modelObject: Offer): DocumentData => ({
+    cavUserRef: modelObject.cavUserRef,
+    pinUserRef: modelObject.pinUserRef,
+    requestRef: modelObject.requestRef,
+    cavUserSnapshot: modelObject.cavUserSnapshot.toObject(),
+    message: modelObject.message,
+    status: modelObject.status,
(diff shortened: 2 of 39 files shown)


firebase/firestore.indexes.json
```diff
@@ -1,4 +1,12 @@
...
+  "indexes": [
+    {
+      "collectionId": "requests",
+      "fields": [
+        { "fieldPath": "status", "mode": "ASCENDING" },
+        { "fieldPath": "latLng", "mode": "ASCENDING" }
+      ]
+    }
+  ],
   "fieldOverrides": []
```
sabind, line 11: Can we also add the code to create the other index we needed (there were two indexes used).

functions/src/models/offers/index.ts
```diff
@@ -1,5 +1,6 @@
+import { FirestoreDataConverter } from '@google-cloud/firestore';
 import { IsEnum, IsNotEmpty, IsObject, IsString, ValidateNested } from 'class-validator';
-import { firestore } from 'firebase';
+import { firestore } from 'firebase-admin';
```
sabind, line 3: Is there a reason for this change?

This code should have been working without this change.

functions/src/models/offers/index.ts
```diff
@@ -149,9 +150,17 @@ export class Offer implements IOffer {
...
-  toFirestore: (modelObject: Offer): DocumentData => modelObject.toObject(),
+  toFirestore: (modelObject: Offer): DocumentData => ({
+    cavUserRef: modelObject.cavUserRef,
+    pinUserRef: modelObject.pinUserRef,
+    requestRef: modelObject.requestRef,
+    cavUserSnapshot: modelObject.cavUserSnapshot.toObject(),
+    message: modelObject.message,
+    status: modelObject.status,
+    createdAt: modelObject.createdAt.toDate(),
+  }),
```
sabind, line 165: Is there any reason we moved away from `model.toObject` here? We don't want to have to maintain this code in two functions.

functions/src/models/offers/offersWithLocation.ts
```diff
@@ -0,0 +1,83 @@
+import { FirestoreDataConverter } from '@google-cloud/firestore';
+import { IsArray } from 'class-validator';
+import { firestore } from 'firebase-admin';
```
sabind, line 3: Given this is a CF only model this is OK, but we want our model classes to use regular firebase so they can later be shared by Web and CF based projects.

functions/src/models/requests/RequestWithOffersAndTimeline.ts
```diff
@@ -0,0 +1,178 @@
+import { FirestoreDataConverter } from '@google-cloud/firestore';
+import { IsArray } from 'class-validator';
+import { firestore } from 'firebase-admin';
```
sabind, line 3: Same comment as before regarding cross-compat with backend/frontend usage.

functions/src/models/requests/index.ts
```diff
@@ -1,6 +1,6 @@
 import { FirestoreDataConverter } from '@google-cloud/firestore';
 import { Allow, IsEnum, IsInt, IsNotEmpty, IsNotEmptyObject, IsObject, IsOptional, IsString, Max, Min, ValidateNested } from 'class-validator';
-import { firestore } from 'firebase';
+import { firestore } from 'firebase-admin';
```
sabind, line 3: Same comment as before regarding cross-compat with backend/frontend usage.

functions/src/models/requests/index.ts
```diff
@@ -299,6 +300,22 @@ export const RequestFirestoreConverter: FirestoreDataConverter<Request> = {
...
+      latLng: JSON.stringify(modelObject.latLng),
+      streetAddress: modelObject.streetAddress,
+      status: modelObject.status,
+      createdAt: modelObject.createdAt.toDate(),
+      updatedAt: modelObject.updatedAt.toDate(),
+      pinRating: modelObject.pinRating,
+      cavRating: modelObject.cavRating,
+      pinRatedAt: modelObject.pinRatedAt?.toDate() || null,
+      cavRatedAt: modelObject.cavRatedAt?.toDate() || null,
+    };
```
sabind, line 319: Is there any reason we moved away from `model.toObject` here? We don't want to have to maintain this code in two functions.

functions/src/models/requests/timeline.ts
```diff
@@ -168,9 +168,9 @@ export class TimelineItem implements ITimelineItem {
 
   toObject(): object {
     return {
-      actorRef: this.actorRef,
-      offerRef: this.offerRef,
-      requestRef: this.requestRef,
+      actorRef: this.actorRef.path,
+      offerRef: this.offerRef?.path,
+      requestRef: this.requestRef.path,
```
sabind, line 173: This is part of the challenge using the toObject function. We have multiple opinions in terms of "what is an object".

I'd say right now an "object" is what we get in and out of firestore, which is how it's been implemented everwhere else.

I'd remove the `.path` here.

sabind:
Resolved some conflicts 👍

firebase/firestore.indexes.json
ashwinkjoseph: that index is no longer being used. that was created initially

functions/src/models/offers/index.ts
ashwinkjoseph: yea, so the `firebase` package is for the client-side, and we are using `firebase-admin` in the cloud function, right? so I thought of keeping it the same as the cloud function.
We can change it back

functions/src/models/offers/index.ts
ashwinkjoseph: yea, this was because we needed to convert the timestamps into Date objects for the tests to pass.
But when I tried the same on the response that I sent to the client, it was behaving strangely (the dates weren't having any values for some reason). I wasn't able to figure out why.
So when it threw an error with the timestamp in the tests, I thought of not making the change in the `.toObject` the method as that's the one we are using in the cloud function and aren't using the converter for the reason I mentioned to you regarding the errors I was getting with the converters in the cloud function and you were using the same method too

functions/src/models/offers/offersWithLocation.ts
ashwinkjoseph: right, but I am not fully sure if we can do that right now, I have the change for client side libraries in the models I used in the `web-client` project.

functions/src/models/requests/index.ts
ashwinkjoseph: https://github.com/reach4help/reach4help/pull/611#discussion_r433959646
apart from what I mentioned in the comment linked here, I also have the issue of latLng, to which I was absolutely not able to figure out any solution at all apart from serializing.
But this will destroy the data structure and the frontend would get values that I can't use.
since I am using the `toObject` method in the cloud function, I didn't want to touch it.

This is purely for fixing the errors in the tests

functions/src/models/requests/timeline.ts
ashwinkjoseph: the reason for doing this was because of how the cloud function was sending the data to the frontend, it was being serialized so the data types were getting destroyed.
which is why if you look at the frontend code, appropriate code has been put to reconstruct the data type. but it's not an ideal solution.
this is only for these abstract models and not for the actual model

PR #612, opened 2020-06-01
Radial Distance function (Haversine Distance)

mcverter:
Radial Distance function (Haversine Distance)


sabind:
This is good, does the code get used anywhere?

PR #615, opened 2020-06-01
Proposed changes to PR

sabind:
## Release Checklist

## Security Checklist

### DIFF
--- a/firebase/firestore.indexes.json
+++ b/firebase/firestore.indexes.json
@@ -6,6 +6,14 @@
         { "fieldPath": "status", "mode": "ASCENDING" },
         { "fieldPath": "latLng", "mode": "ASCENDING" }
       ]
+    },
+    {
+      "collectionId": "requests",
+      "fields": [
+        { "fieldPath": "pinUserRef", "mode": "ASCENDING" },
+        { "fieldPath": "status", "mode": "ASCENDING" },
+        { "fieldPath": "latLng", "mode": "ASCENDING" }
+      ]
     }
   ],
   "fieldOverrides": []
--- a/firebase/firestore.indexes.json
+++ b/firebase/firestore.indexes.json
@@ -1,4 +1,12 @@
 {
-  "indexes": [],
+  "indexes": [
+    {
+      "collectionId": "requests",
+      "fields": [
+        { "fieldPath": "status", "mode": "ASCENDING" },
+        { "fieldPath": "latLng", "mode": "ASCENDING" }
+      ]
+    }
+  ],
   "fieldOverrides": []
 }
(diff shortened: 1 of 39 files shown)


firebase/firestore.indexes.json
```diff
@@ -6,6 +6,14 @@
         { "fieldPath": "status", "mode": "ASCENDING" },
         { "fieldPath": "latLng", "mode": "ASCENDING" }
       ]
+    },
+    {
+      "collectionId": "requests",
```
ashwinkjoseph, line 11: these aren't being used by the current queries. should we still keep them?

functions/src/models/offers/index.ts
```diff
@@ -1,8 +1,8 @@
 import { FirestoreDataConverter } from '@google-cloud/firestore';
 import { IsEnum, IsNotEmpty, IsObject, IsString, ValidateNested } from 'class-validator';
-import { firestore } from 'firebase-admin';
+import { firestore } from 'firebase';
```
ashwinkjoseph, line 3: This was my first approach at solving the lint errors.
But because we are using the `firestore` instance from `firebase-admin`, I am not sure if this would cause issues in the cloud function? maybe we would have to use `Type Generics` here and have a function that would use appropriate type based on if it's a test and whether the test is demanding client or server. I think that would be a better solution. but for now wouldn't the existing solution be enough

functions/src/models/offers/offersWithLocation.ts
```diff
@@ -61,9 +60,9 @@ export class OfferWithLocation extends Offer implements IOfferWithLocation {
 
   public toObject(): object {
     return {
-      cavUserRef: this.cavUserRef.path,
-      pinUserRef: this.pinUserRef.path,
-      requestRef: this.requestRef.path,
+      cavUserRef: this.cavUserRef,
```
ashwinkjoseph, line 63: have you tried running the `web-client` with these changes?

functions/src/models/requests/index.ts
```diff
@@ -300,22 +312,14 @@ export const RequestFirestoreConverter: FirestoreDataConverter<Request> = {
...
-      streetAddress: modelObject.streetAddress,
-      status: modelObject.status,
-      createdAt: modelObject.createdAt.toDate(),
-      updatedAt: modelObject.updatedAt.toDate(),
-      pinRating: modelObject.pinRating,
-      cavRating: modelObject.cavRating,
-      pinRatedAt: modelObject.pinRatedAt?.toDate() || null,
-      cavRatedAt: modelObject.cavRatedAt?.toDate() || null,
-    };
+    const object = modelObject.toObject();
```
ashwinkjoseph, line 315: have you tried running the `tests` with these changes? they should be throwing errors due to the Timestamp.
but it could very well be an issue with the types

PR #630, opened 2020-06-03
Inserts actual distance between Request and each Offer in OfferList

mcverter:


mcverter:
offerlist

mcverter:
offerlist

mcverter:
@ashwinkjoseph I have not yet tested -- I don't have two separate users, one for offer and one for request.  I will try to create these and test.  You can look over the code and test on your own environment if you want

web-client/src/modules/timeline/components/OffersList/OffersList.tsx
```diff
@@ -103,6 +106,7 @@ const ButtonsContainer = styled.div`
 const OfferItem: React.FC<OfferItemProps> = ({
   offer,
   handleOffer,
+  requestCoords,
```
ashwinkjoseph, line 109: I see that in the timeline container, you have passed this as a prop to the offersList and the offersList component isn't passing it down to the offerItem anywhere

web-client/src/modules/timeline/containers/TimelineViewContainer/TimelineViewContainer.tsx
```diff
@@ -43,6 +43,15 @@ const TimelineViewContainer: React.FC<TimelineViewContainerProps> = ({
     ({ profile }: { profile: ProfileState }) => profile,
   );
 
+  const requestCoords = useSelector(({ profile }: { profile: ProfileState }) =>
+    profile &&
+    profile.privilegedInformation &&
+    profile.privilegedInformation.address &&
+    profile.privilegedInformation.address.coords
+      ? profile.privilegedInformation?.address.coords
+      : { latitude: 0, longitude: 0 },
```
ashwinkjoseph, line 52: it would be better to use the `GeoPoint` constructor here, but this works too

web-client/src/modules/timeline/components/OffersList/OffersList.tsx
```diff
@@ -128,8 +132,9 @@ const OfferItem: React.FC<OfferItemProps> = ({
       <IconContainer>
         <EnvironmentOutlined />
       </IconContainer>
-      {/* TODO : Logic for using offer.address.coords for calculating the distance */}
-      <TextIcon>5 KM</TextIcon>
+      <TextIcon>
+        {haversineDistance(data.address.coords, requestCoords)}
```
ashwinkjoseph, line 136: where do you get data from? did you intend for this to be `offer.address.coords`?

web-client/src/modules/timeline/components/OffersList/OffersList.tsx
```diff
@@ -18,6 +20,7 @@ interface OffersListProps {
 interface OfferItemProps {
   offer: Offer;
```
ashwinkjoseph, line 21: the data type of offer here should be changed to `OfferWithLocation` (the class abstract class from the backend which combines the details of the offer and the location/address).
This same change should happen in the `OffersListProps` in line 15 as well. there it should be `Record<string, OfferWithLocaiton>`

ashwinkjoseph:
> @ashwinkjoseph I have not yet tested -- I don't have two separate users, one for offer and one for request. I will try to create these and test. You can look over the code and test on your own environment if you want

Since you would be testing on the `reach4help-dev` firebase project, you should be able to switch between CAV and PIN and be able to make and accept offers with the same user.
This is because I disabled the rule that prevents this in `reach4help-dev` for the purposes of faster development

sabind:
> This is because I disabled the rule that prevents this in `reach4help-dev` for the purposes of faster development

How did you do that ?

ashwinkjoseph:
https://github.com/reach4help/reach4help/pull/630#issuecomment-638947675
😅 I just edited it from the console for the dev project. 
since it was on the dev, I figured it wouldn't be a problem

web-client/src/modules/timeline/components/OffersList/OffersList.tsx
ashwinkjoseph: oh, my bad.

Issue #644, opened 2020-06-05
Change wording on Timeline "PIN created" => "Person In Need created" or "You created"


mcverter:
**Describe the bug**
Timeline  refers to user as "PIN" which is our internal name for the state.
Should be "Person In Need created" if PIN === OtherUser
Should be "You created" if PIN === CurrentUser

Check for other usages of "PIN"

**To Reproduce**
![PINcreated](https://user-images.githubusercontent.com/1453956/83892603-e3bda180-a71c-11ea-925d-a2a9e73cfc44.png)

1.  Create a request as PIN
2.  switch to CAV
3.  Click on any Open Request

**Expected behavior**
"PIN" should be replaced with "Person In Need created" if OtherUser
"PIN" should be replaced with "You created" if CurrentUser

Try to replace all instances of "PIN"

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/644)

*This information is updated automatically. To modify it, please use ZenHub.*

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/644)

*This information is updated automatically. To modify it, please use ZenHub.*

mcverter:
![timelineNames](https://user-images.githubusercontent.com/1453956/84079543-c00e8b80-a9a8-11ea-97cf-9267ef5c54cd.png)

Figma shows we want "You" and username.
I have implemented a temporary fix "Person In Need" and "Volunteer"

@btuerker can you fix the logic?

PhilArsHermit:
Verified in R4H v2.1.0.  Fixed

PIN
![PIN_Timeline](https://user-images.githubusercontent.com/65713824/86484221-c08a1000-bd23-11ea-81c2-918a28c1b5e9.png)

Volunteer
![Volunteer_Timeline](https://user-images.githubusercontent.com/65713824/86484237-c67ff100-bd23-11ea-9194-af05c5bb411a.png)


Closing the ticket

PR #651, opened 2020-06-07
Bug/644 fix timeline wording

mcverter:
fixes wording on timeline
From "PIN" to "Person In Need"
From "CAV" to "Volunteer"
https://github.com/reach4help/reach4help/issues/644

PINcaV


mcverter:
Note that there is a LOGIC ERROR in the screenshot above:  "Volunteer accepted this request" is on PIN side.  Can we make a new issue and maybe @btuerker can look into it?

ashwinkjoseph:
https://github.com/reach4help/reach4help/pull/651#issuecomment-640240363

This is actually in an issue with the backend. I will look into it and fix it now

Issue #647, opened 2020-06-05
[BUG][WCAG] requests/new - Little to no feedback on buttons to indicate focus. 


j-isenberg:
**Describe the bug**
When the user hovers over or tabs into certain buttons on the requests/new page, there is either not enough of a change to the button or no change at all to indicate focus. 
All elements with user interaction must indicate when it is in focus, whether by hovering over it with the cursor, or tabbing into the element with the keyboard. 

**1. Use my location:** The colour does not change enough to be considered compliant when the user hovers over or tabs into this button. 
Idle hex code: 7D00A3
Focused hex code: A12596
Minimum contrast requirement: 3:1

**2. "Menu", "New request" and "Notification" buttons:** There is NO feedback on any of these buttons. 

**Optional. "Search" and "Cancel" button text:** The colour of the text does not have a very high contrast when hovered over. This is optional because the boarder of these buttons do change enough. 

https://reach4help-staging-web-client.firebaseapp.com/requests/new

**To Reproduce**
1. Navigate to requests/new page. 
2. Hover over each button.
3 (w/ Axe). Check hex codes of idle and focused buttons. 
4. Notice that contrasts do not meet minimum requirements. 
5. Notice that focusing over the menu, new request and notification buttons does nothing. 

**Expected behavior**
Users are given proper feedback to indicate focus. Feedback include compliant contrast ratios. 

**Screenshots**
Video posted in Slack team-user-testing channel. 
https://slack-files.com/T010H81FF60-F015MQLBQ48-620cd65262

**Desktop (please complete the following information):**
- Windows 10 - Chrome 

**Additional context**
Two-color focus indicator techniques: https://www.w3.org/WAI/WCAG21/Techniques/css/C40
https://contrastchecker.com/ 
https://www.w3.org/WAI/WCAG21/quickref/#focus-visible

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/647)

*This information is updated automatically. To modify it, please use ZenHub.*

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/647)

*This information is updated automatically. To modify it, please use ZenHub.*

j-isenberg:
Most buttons on the Request/new page now gives the user feedback when hovered over or tabbed into. 
[...]

Issue #638, opened 2020-06-04
Buttons reach goes all the way to the right of the page


PhilArsHermit:
**Describe the bug**
All the buttons in the "Settings" menu have a reach that goes all the way to the right of the page.  Meaning that the user can inadvertently click on the button even if the cursor is not on it.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to 'https://reach4help-staging-web-client.firebaseapp.com/requests/find'
2. Click on the three horizontal bar
3. Get the cursor on any button
- Find Requests
- Open
- Accepted
- Ongoing
- Finished
- Archived
- Contact Us
- Switch to Volunteer
- Sign Out
4. Pass the cursor through the right

**Expected behavior**
The "hand" cursor is still showing, meaning that the user if the user clicks, it will trigger the button even if the cursor is no longer on the button.

**Screenshots**
![Button_Reach](https://user-images.githubusercontent.com/65713824/83777034-ece43b00-a656-11ea-9f6b-b6d9afe79c60.png)

**Desktop (please complete the following information):**
Build 1
 - OS: W10 Pro
 - Browser Chrome
 - Version 83.0.4103.97

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/638)

*This information is updated automatically. To modify it, please use ZenHub.*

ashwinkjoseph:
I agree. This should be fixed by having different styles for when the user highlights or hovers over a button

isma-osy:
Winggo

CorporateInvesther:
Verified Fixed v2.1.0

PR #653, opened 2020-06-07
Bug/647 add hover to buttons

mcverter:
fixes these two issues:
https://github.com/reach4help/reach4help/issues/647
https://github.com/reach4help/reach4help/issues/638

Adds hover to side menu, bottom menu, and New Request screen

### DIFF
--- a/web-client/src/components/Buttons/index.tsx
+++ b/web-client/src/components/Buttons/index.tsx
@@ -1,0 +1,31 @@
+import { Button } from 'antd';
+import { COLORS } from 'src/theme/colors';
+import styled from 'styled-components';
+
+export const StepBackButton = styled(Button)`
+  border-radius: 4px;
+  width: 100%;
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
+  &:hover {
+    background-color: ${COLORS.backgroundAlternative};
+    color: white;
+    font-weight: 700;
+  }
+`;
+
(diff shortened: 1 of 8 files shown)


mcverter:
@sabind @ashwinkjoseph  After the MVP, we need to do a refactor of the code.
This starts to solve the problem by establishing standard buttons in the src/components directory.  
I left some code for PrimaryButton intact because it looked promising even though it wasn't working.

Note that there is one button that does not produce the same hover effect -- when you click on a marker on the Map and it says "Do you want to help <PIN>".  This does have a hover effect, but not the same as the others.  I didn't spend time adjusting this because it's okay for now and we need to do a bigger refactor of the code.

web-client/src/modules/timeline/components/BottomPanel/BottomPanel.tsx
```diff
@@ -192,34 +196,32 @@ const ButtonContainer = styled(FlexDiv)`
     width: 100%;
   }
 `;
+/* TODO:  Check if there is any useful code here
```
mcverter, line 199: I left this code in the file to review later.  It looks like it should do something interesting but is not currently working

web-client/src/components/Buttons/index.tsx
```diff
@@ -0,0 +1,31 @@
+import { Button } from 'antd';
```
mcverter, line 1: I support the creation of a reusable component library.  This buttons file is a start.

Issue #662, opened 2020-06-08
New Phone Screen


mcverter:
![newTelephoneScreen](https://user-images.githubusercontent.com/1453956/84076214-6192de80-a9a3-11ea-93f6-6bad82716996.png)

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/662)

*This information is updated automatically. To modify it, please use ZenHub.*

PR #671, opened 2020-06-10
Bug/662 change phone entry screen

mcverter:
https://github.com/reach4help/reach4help/issues/662

Redesign Phone Entry screen with separate dropdown input for country code.
See screen shot on bug for details


mcverter:
Also some small UI cleanup

PR #687, opened 2020-06-11
Fix/create request screen css issues

btuerker:
## Description
Fixes full-screen map issue on the safari browser also improves scrolling on touch screens. I've tested this improvements on my phone as well. Scrolling works perfectly.

Also, I've added some meta keys to the HTML file. Now, it's not zooming in when the keyboard is opened for filling inputs.

The way I made the map responsive might not look perfect but the other solutions caused terrible scrolling experience. However, we can easily overwrite it by writing media queries for the desktop version.

## Release Checklist

## Security Checklist

## Additional Notes

Screen Shot 2020-06-11 at 1 12 29 AM

Screen Shot 2020-06-11 at 1 12 34 AM

Screen Shot 2020-06-11 at 1 12 22 AM

Screen Shot 2020-06-11 at 1 37 10 AM

Screen Shot 2020-06-11 at 1 37 33 AM

Screen Shot 2020-06-11 at 1 37 24 AM


sabind:
We should always show the form. That's the most critical path. Lets get that fixed in this PR and merge it once it's working.

web-client/src/modules/requests/containers/NewRequestsContainer/NewRequestsContainer.tsx
```diff
@@ -29,6 +29,12 @@ const RequestDetails = styled.div`
   background: white;
 `;
 
+const MapContainer = styled.div`
```
mcverter, line 32: @btuerker It would be nicer if we could use a media query rather than giving it a specific height
https://www.w3schools.com/cssref/css3_pr_mediaquery.asp

I had played around with the proper height of the map.  The problem is that the NewRequest form takes up so much space on smaller devices that not enough is left for the map -- at least on the screens provided by my google chrome responsive mode.   

The css in the NewRequestForm needs to be tighter.  It takes up too much height.

We solved this by wrapping the NewRequestForm and the Map in a flex container.

I had had performance issues with heights < 100%:  Either the center marker would disappear below the UI or gray bars would appear around the map when the user increased or decreased the zoom of the browser itself -- the map wouldn't fill the entirety of its HTML box.

That said, I am okay for now giving this component a height less than 100% to resolve the urgent bug, but we want to test for any bugs this might create, and we may want to wrap this in a media query for different sized screens

web-client/src/components/DashboardLayout/DashboardLayout.tsx
```diff
@@ -22,7 +23,7 @@ const DashboardLayout: React.FC<DashboardLayoutProps> = ({
   const [notificationVisible, setNotificationVisible] = useState(false);
 
   return (
-    <Layout>
+    <StyledLayout>
```
mcverter, line 26: Use a descriptive name.  Styled for what purpose?  What is this component doing?  Treat it like any other React component and give it a real name.

web-client/src/components/DashboardLayout/DashboardLayout.tsx
```diff
@@ -39,20 +40,25 @@ const DashboardLayout: React.FC<DashboardLayoutProps> = ({
...
         openMenu={() => setMenuVisible(true)}
         openNotifications={() => setNotificationVisible(true)}
         isCav={isCav}
       />
-    </Layout>
+    </StyledLayout>
   );
 };
 
+const StyledLayout = styled(Layout)`
```
mcverter, line 53: Change name

web-client/src/components/DashboardLayout/DashboardLayout.tsx
```diff
@@ -39,20 +40,25 @@ const DashboardLayout: React.FC<DashboardLayoutProps> = ({
...
+  height: 100vh;
+`;
+
+const StyledLayoutContent = styled(Layout.Content)`
+  margin: 64px 0;
```
mcverter, line 58: Change name

web-client/src/components/DashboardLayout/DashboardLayout.tsx
btuerker: I considered this, but `Layout` is Antd component so, I approached here to indicate this is the styled version of Antd `Layout` component. The reason is, we'll need to write some media queries for the desktop layout.

web-client/src/modules/requests/containers/NewRequestsContainer/NewRequestsContainer.tsx
btuerker: > @sabind I would like your opinion on this.
> This fixes the issue on mobile displays and now it shows up perfectly fine on mobile.
> But it doesn't show the request form anywhere in the desktop view, not even when we scroll the page
> image
> 
> I would like to request @btuerker for fixing this, but should we accept this now and fix it later? because mobile is our top priority right? so we could deploy this now and @btuerker could work on fixing this issue on the desktop by tomorrow or he could fix it and we will merge it tomorrow along with the fix

max-height would solve this, for now, I'm considering that we'll entirely change this screen on desktop view. Figma designs show that desktop view uses modal for request create operation. Also, I can put a max-width to the parent component and aligned it to the center. I say that again, just for this moment. Here's how it would look:

Screen Shot 2020-06-11 at 6 39 15 PM

web-client/src/components/DashboardLayout/DashboardLayout.tsx
mcverter: It is very hard to work with and refactor code if you use generic names like "Layout"
This is a ReactComponent and should be named like any other component
If you wan to indicate inheritance relationship with antd.Layout, call it "MapDisplayLayout"

web-client/src/modules/requests/containers/NewRequestsContainer/NewRequestsContainer.tsx
mcverter: I think adding @media queries would help to solve this problem

sabind: @mcverter @btuerker It' sounds like we're making some progress here. Let's see if we can't improve this, but I like the progress we've made. I don't honestly know if mobile or desktop will be more popular. I think it depends on age. For example PINs may be more desktop based (or at least iPad).

mcverter: @sabind @btuerker just add a media query on line 33 to adjust

btuerker: @sabind I'm planning to improve the desktop layout for the later versions. At least, we should remove the bottom navbar and show the sider menu on left.

However, @sabind can I just fix the desktop view like below: 
Screen Shot 2020-06-11 at 7 19 49 PM

web-client/src/components/DashboardLayout/DashboardLayout.tsx
btuerker: This is in the `DashboardLayout` component, not in the request module, or isn't only related to this screen, so `MapDisplayLayout` is not an option here. IMO, this is the only place we can use some generic names as it's responsible for the entire layout system.

Issue #672, opened 2020-06-10
Use only Template Strings: Prepare front end for Portuguese and other languages


mcverter:
Currently many strings in project are raw text rather than using translation object

Wrong
const foo = “bar”

Right 
const foo = t(‘strings.bar’)

Must be fixed throughout code and JSX markup

To Test you need to do a thorough review of everything.  Code compilation will not detect this type of error:
Oops -- I misspelled the name of the template
`
<Button>{ t.strings.bra }</Button>
`

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/672)

*This information is updated automatically. To modify it, please use ZenHub.*

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/672)

*This information is updated automatically. To modify it, please use ZenHub.*

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/672)

*This information is updated automatically. To modify it, please use ZenHub.*

mcverter:
this issue is too general.  it has already been resolved, and does not point to a current real product need.
Closing issue

PR #689, opened 2020-06-11
Bug/672 only template strings

mcverter:
https://github.com/reach4help/reach4help/issues/672

All text in project, either in code or in JSX markup, must use international template strings so that we can proceed with translation
Wrong
const foo = “bar”

Right
const foo = t(‘strings.bar’)

QA should make sure that all the template strings properly compiled throughout the app.

If they see "I had a drink at the strings.bar" instead of "I had a drink at the bar", this means we need to fix a template strings.


mcverter:
@ashwinkjoseph @sabind 
Use
https://github.com/reach4help/reach4help/pull/689/files?w=1
to ignore whitespace differences

mcverter:
@btuerker I saw that you recently worked on this pull to add interpolation to template strings.
https://github.com/reach4help/reach4help/pull/686

Maybe we can go over template strings in the codebase after mvp

mcverter:
stringInterpolationError
Here is an example of a string interpolation error

PR #713, opened 2020-06-12
Bug/prevent google manage address

mcverter:
manageAddressError


mcverter:
all language about "Country" has been removed.  google autodetects this and imposes the forms

PR #735, opened 2020-06-12
BottomNavbar has boolean for visibility

mcverter:
## Release Checklist

## Security Checklist


mcverter:
this was at @sabind 's request.  
he wanted the bottom buttons to be hidden when we are looking at notifications.  
even if this isn't wanted, i think putting a visibility boolean on the bottom buttons is a good idea and it is easy to implement

ashwinkjoseph:
> this was at @sabind 's request.
> he wanted the bottom buttons to be hidden when we are looking at notifications.
> even if this isn't wanted, i think putting a visibility boolean on the bottom buttons is a good idea and it is easy to implement

I think a better approach to achieving what @sabind wanted was to modify the notifications drawer component to fill the whole page like the sidebar.

The reason why I don't think it's a good idea to have a visibility attribute in the bottom panel is explained here: https://github.com/reach4help/reach4help/pull/735#pullrequestreview-430639038

sabind:
> this was at @sabind 's request.
> he wanted the bottom buttons to be hidden when we are looking at notifications.
> even if this isn't wanted, i think putting a visibility boolean on the bottom buttons is a good idea and it is easy to implement

If you open the "Notifications" you'll see a really weird issue with the bottom nav click handlers and the notifications side staying open. I mentioned it should cover the whole screen and it actually seems this will be very helpful to enhance the bottom bar with this (and solve our problem)

ashwinkjoseph:
> > this was at @sabind 's request.
> > he wanted the bottom buttons to be hidden when we are looking at notifications.
> > even if this isn't wanted, i think putting a visibility boolean on the bottom buttons is a good idea and it is easy to implement
> 
> If you open the "Notifications" you'll see a really weird issue with the bottom nav click handlers and the notifications side staying open. I mentioned it should cover the whole screen and it actually seems this will be very helpful to enhance the bottom bar with this (and solve our problem)

well, my idea was to modify the notification drawer. 
but this works if it works.

but @mcverter discuss this with @btuerker then.

one thing both of you have to be mindful of is how the `MasterPage.ts` decides which layout to use based on the definition inside the module. If this solution adds any value to that implementation then I think this is good to go

mcverter:
@s

> > > this was at @sabind 's request.
> > > he wanted the bottom buttons to be hidden when we are looking at notifications.
> > > even if this isn't wanted, i think putting a visibility boolean on the bottom buttons is a good idea and it is easy to implement
> > 
> > 
> > If you open the "Notifications" you'll see a really weird issue with the bottom nav click handlers and the notifications side staying open. I mentioned it should cover the whole screen and it actually seems this will be very helpful to enhance the bottom bar with this (and solve our problem)
> 
> well, my idea was to modify the notification drawer.
> but this works if it works.
> 
> but @mcverter discuss this with @btuerker then.
> 
> one thing both of you have to be mindful of is how the `MasterPage.ts` decides which layout to use based on the definition inside the module. If this solution adds any value to that implementation then I think this is good to go

Before this change, the BottomButtons were always open, but the MenuDrawer was covering them.  You could still see them in the DOM but they were beneath other elements.  

For a container page like MasterPage.tsx, it is often good to set booleans on elements that can appear or disappear in different context.  By doing this, burhan can now use the entirety of the page for the notifications.

mcverter:
> one thing both of you have to be mindful of is how the `MasterPage.ts` decides which layout to use based on the definition inside the module.

This sounds like "Dependency Inversion"
https://en.wikipedia.org/wiki/Dependency_inversion_principle

A higher container element like MasterPage.ts should never depend on its children to determine its UI

ashwinkjoseph:
> > one thing both of you have to be mindful of is how the `MasterPage.ts` decides which layout to use based on the definition inside the module.
> 
> This sounds like "Dependency Inversion"
> https://en.wikipedia.org/wiki/Dependency_inversion_principle
> 
> A higher container element like MasterPage.ts should never depend on its children to determine its UI

yea, but if we do this and make sense of this change for the future, it has to depend on the children.

but with the current implementation, it's not like that. well. I guess we can discuss this on the daily stand up call.

also, this isn't a big issue. you may have issues making sense of the code later in the future and someone might blame the codebase. But it wouldn't break the project in any way

mcverter:
> > > one thing both of you have to be mindful of is how the `MasterPage.ts` decides which layout to use based on the definition inside the module.
> > 
> > 
> > This sounds like "Dependency Inversion"
> > https://en.wikipedia.org/wiki/Dependency_inversion_principle
> > A higher container element like MasterPage.ts should never depend on its children to determine its UI
> 
> yea, but if we do this and make sense of this change for the future, it has to depend on the children.
> 
> but with the current implementation, it's not like that. well. I guess we can discuss this on the daily stand up call.
> 
> also, this isn't a big issue. you may have issues making sense of the code later in the future and someone might blame the codebase. But it wouldn't break the project in any way

Why does it have to "depend on the children"?  This is an antipattern and will make the code harder to maintain.  We need to discuss these changes before anyone goes down this path

ashwinkjoseph:
yup

> > > > one thing both of you have to be mindful of is how the `MasterPage.ts` decides which layout to use based on the definition inside the module.
> > > 
> > > 
> > > This sounds like "Dependency Inversion"
> > > https://en.wikipedia.org/wiki/Dependency_inversion_principle
> > > A higher container element like MasterPage.ts should never depend on its children to determine its UI
> > 
> > 
> > yea, but if we do this and make sense of this change for the future, it has to depend on the children.
> > but with the current implementation, it's not like that. well. I guess we can discuss this on the daily stand up call.
> > also, this isn't a big issue. you may have issues making sense of the code later in the future and someone might blame the codebase. But it wouldn't break the project in any way
> 
> Why does it have to "depend on the children"? This is an antipattern and will make the code harder to maintain. We need to discuss these changes before anyone goes down this path

yes, you are right. what I am saying is that this change could lead to a situation like that in the future. 

but in order to understand my concern fully, I think I have to explain the MaterPage.ts to you and also the layouts and I think the frontend pod/standup call would be the perfect time for me to do this

ashwinkjoseph:
z

PR #739, opened 2020-06-13
changing roles from nouns (permanent identities) to verbs (dynamic ac…

mcverter:
Changes language from PIN / CAV -- nouns that convey permanent activities to verbs: looking, supporting, helping, etc.

### DIFF
--- a/web-client/src/translations/en.json
+++ b/web-client/src/translations/en.json
@@ -38,8 +38,8 @@
     "menuDrawer": {
       "contactUs": "Contact us",
       "logout": "Sign out",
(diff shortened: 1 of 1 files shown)


schowdhury8:
Look great, but maybe instead of "Offering Support" just "Helping"? 
"Helping" vs "Seeking Help" seems like an easier 1:1 comparison to me, but you can't go wrong with this either.

sabind:
"Providing Help"
"Seeking Help"


If we want to be consistent ^

sabind:
That being said, the general concept behind this is awesome. Thank you @mcverter for taking the lead here.

Issue #663, opened 2020-06-08
Address Errors are all the same


mcverter:
![addressErrors](https://user-images.githubusercontent.com/1453956/84086493-04ecef00-a9b6-11ea-8afc-ddfd8040a00b.png)

Should be different messages for each field

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/663)

*This information is updated automatically. To modify it, please use ZenHub.*

PR #787, opened 2020-06-22
Fix PersonalDataFrom fields error messages

btuerker:
## Description

- I added specific error messages for the city and country form fields.

- I added interpolation to terms & condition field to prevent possible translation issues.

- I also refactored the language file, but it will cause us to rewrite all the translations made before. If we don't want to deal with it right a moment, I'll remove those changes.

## Motivation
#663 
## Release Checklist

## Security Checklist

#### Here're the fields changed: 
Screen Shot 2020-06-22 at 10 39 33 AM

#### The English version of the Terms & Conditions field: 
Screen Shot 2020-06-22 at 11 26 02 AM
#### The Turkish version of Terms & Conditions field: (just a mock(didn't manipulate the Turkish language file) to show the order of words can be changed)
Screen Shot 2020-06-22 at 11 25 06 AM

### DIFF
--- a/web-client/src/modules/personalData/components/PersonalDataForm/PersonalDataForm.tsx
+++ b/web-client/src/modules/personalData/components/PersonalDataForm/PersonalDataForm.tsx
@@ -11,7 +11,7 @@ import {
 import { firestore } from 'firebase';
 import words from 'lodash/words';
 import React, { Fragment, useEffect, useState } from 'react';
-import { useTranslation } from 'react-i18next';
+import { Trans, useTranslation } from 'react-i18next';
 import { DEVICE_MIN } from 'src/constants/mediaQueries';
 import { User } from 'src/models/users';
 import {
@@ -286,7 +286,9 @@ const PersonalDataForm: React.FC<PersonalDataFormProps> = ({
   return (
     <Introduction className="withContentPaddingDesktop">
       {displayPic && <ProfilePhoto src={displayPic} />}
-      <Title>{t('user_data_form.sub_title')}</Title>
+      <Title>
+        {t('modules.personal-data.components.PersonalDataForm.sub_title')}
+      </Title>
       <Form
         layout="vertical"
         form={form}
@@ -301,7 +303,9 @@ const PersonalDataForm: React.FC<PersonalDataFormProps> = ({
               rules={[
                 {
                   required: true,
-                  message: t('user_data_form.full_name_error_message'),
+                  message: t(
+                    'modules.personal-data.components.PersonalDataForm.full_name_error_message',
+                  ),
                 },
               ]}
               label={t('name')}
@@ -322,7 +326,9 @@ const PersonalDataForm: React.FC<PersonalDataFormProps> = ({
               rules={[
                 {
                   required: true,
-                  message: t('user_data_form.display_name_error_message'),
+                  message: t(
+                    'modules.personal-data.components.PersonalDataForm.display_name_error_message',
+                  ),
                 },
               ]}
             >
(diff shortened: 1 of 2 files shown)


web-client/src/translations/en.json
```diff
@@ -121,31 +121,6 @@
...
-      "policy_link": "Privacy & Security Policy.",
-      "accept_to_use_profile_pic": "I am okay with letting Reach4Help use my profile picture.",
-      "terms_conditions_text": "I understand and acknowledge the ",
-      "terms_conditions_link": "Terms & Conditions.",
-      "terms_conditions_error": "You need to accept the Terms & Conditions.",
-      "full_name_error_message": "Please enter your name.",
-      "display_name_error_message": "Please enter your display name.",
-      "address_error_message": "Please enter your address.",
-      "allow_send_notifications": "I would like Reach4Help to send me notifications and emails."
-    },
```
btuerker, line 148: @mcverter do we use this anywhere else? I checked but didn't see, so I guess it's safe.

web-client/src/modules/personalData/components/PersonalDataForm/PersonalDataForm.tsx
```diff
@@ -497,14 +519,21 @@ const PersonalDataForm: React.FC<PersonalDataFormProps> = ({
...
               >
-                {t('user_data_form.terms_conditions_link')}
-              </a>
+                {/* eslint-disable-next-line jsx-a11y/anchor-has-content,jsx-a11y/control-has-associated-label */}
+                <a
+                  target="_blank"
+                  rel="noopener noreferrer"
+                  href="https://github.com/reach4help/reach4help/blob/master/CODE_OF_CONDUCT.md"
+                />
+              </Trans>
```
btuerker, line 536: @mcverter There's a couple of similar issues within the current i18n file. You can refer to this whenever you need to do complicated interpolations similarly. We should always consider that word orders can be changed from language to another language.

web-client/src/modules/personalData/components/PersonalDataForm/PersonalDataForm.tsx
```diff
@@ -497,14 +519,21 @@ const PersonalDataForm: React.FC<PersonalDataFormProps> = ({
...
+                  ),
+                }}
               >
-                {t('user_data_form.terms_conditions_link')}
-              </a>
+                {/* eslint-disable-next-line jsx-a11y/anchor-has-content,jsx-a11y/control-has-associated-label */}
+                <a
+                  target="_blank"
+                  rel="noopener noreferrer"
+                  href="https://github.com/reach4help/reach4help/blob/master/CODE_OF_CONDUCT.md"
```
btuerker, line 534: We're still referring to the CODE_OF_CONDUCT here, what is the actual link of our terms&conditions page? We can fix it too within this PR.

mcverter: https://github.com/mcverter/reach4help/blob/feature/add-coding-conventions/web-client/docs/CodingConventions/CODING_CONVENTIONS.md

for now only on my branch @btuerker

ashwinkjoseph: we don't have this ready yet. so ignore this for the time being

web-client/src/translations/en.json
ashwinkjoseph: I just noticed that we were using it in the the `UnsupportedLocations.tsx`.
However, it isn't a big deal, that page hasn't been integrated yet and the dev integrating it would probably know that the key is missing when they start integrating the page in the future

PR #816, opened 2020-06-26
base text styles from figma

mcverter:
@sabind @ashwinkjoseph 
This work is preliminary.
Do you think this is a good way to proceed?

I don't know whether predefined BlockStyles will be that useful.
Also, Headers are block-level rather than inline styles, so I am using "H1Text" instead of "H1" as names.

The idea is to start making people do:

< BlockStyle >
   < TextStyle >
     content
  < / TextStyle >
< / BlockStyle >

### DIFF
--- a/web-client/src/components/figma/BlockStyles.ts
+++ b/web-client/src/components/figma/BlockStyles.ts
@@ -1,0 +1,9 @@
+/* Common layout patterns used by Figma */
+
+import styled from 'styled-components';
+
+const ExampleDivStyle = styled.div`
+  margin: 15px;
+`;
+
+export { ExampleDivStyle };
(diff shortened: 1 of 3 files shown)


ashwinkjoseph:
> @sabind @ashwinkjoseph
> This work is preliminary.
> Do you think this is a good way to proceed?
> 
> I don't know whether predefined BlockStyles will be that useful.
> Also, Headers are block-level rather than inline styles, so I am using "H1Text" instead of "H1" as names.
> 
> The idea is to start making people do:
> 
> < BlockStyle >
> < TextStyle >
> content
> < / TextStyle >
> < / BlockStyle >

I actually like this approach, this will be our own asset library, will make the UI elements more uniform across our product.
what do you think @sabind  @puzzledbytheweb ?

Issue #855, opened 2020-07-06
Update CONTRIBUTING.md to include forking/PRing clearly


schowdhury8:
**Describe the feature**
We should make our guidelines for contributing to the project (forking, making a pull request, etc.) a bit more clearly. 
https://github.com/reach4help/reach4help/blob/master/CONTRIBUTING.md#your-first-contribution

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/855)

*This information is updated automatically. To modify it, please use ZenHub.*

mcverter:
@schowdhury8 
This is a bit confusing because we have two CONTRIBUTING.md files. 
I don't think we need the one at the top level because we have one at web-client/ subdirectory
Can we get rid of the top level file or reduce it significantly to just a link to web-cient/Contributing.md

PR #863, opened 2020-07-08
Documentation/855 forking pr instructions

mcverter:
https://github.com/reach4help/reach4help/issues/855

@schowdhury8 @sabind
This is a bit confusing because we have two CONTRIBUTING.md files.
I don't think we need the one at the top level because we have one at web-client/ subdirectory
Can we get rid of the top level file or reduce it significantly to just a link to web-cient/Contributing.md

I am going to make these changes in web-client/CONTRIBUTING.md and recommend that top-level is edited to just a link to web-client for now


schowdhury8:
Since this is a monorepo, I would suggest keeping general contributing guidelines (forking, PRing, etc.) in the top level CONTRIBUTING.md file and keep things specific to a single project (i.e. env vars) in the lower level ones.

PR #889, opened 2020-07-26
Feature/755 testing

skspade:
My time has been limited this week for test writing, but I went ahead and finished `RegistrationSteps` test.

So I never worked with Redux before, initially, I was providing the tests an instance of the Redux store. The more I learned about Redux, the more I realized mocking the whole store is a bad idea. 
Why? 
Because if you are testing a component that relies heavily on redux you probably need to reduce the scope of your test to a child component and just mock out the props. If that is difficult you may want to consider refactoring your component to make dependency injection easier. 

With that being said, I removed the `LoginRoute` test file and reduced the scope of the test that was in there to the `RegistrationSteps` component level.

I also **fixed a bug** for the validator of `Password` and `Confirm Password`. Before the validator would never pass because it was checking the wrong parameter.

I also added a `setupTests` file. This is for all of our global mocks to run tests setups.
For examples:
https://jestjs.io/docs/en/setup-teardown


schowdhury8:
Thanks so much, Sean! To make sure you're appropriately credited, @all-contributors please add @skspade for code and testing!

schowdhury8:
Let's try that again... @all-contributors please add @skspade for code and testing!

schowdhury8:
And again cuz we messed the original PR up... 🤣 
@all-contributors please add @skspade for code and testing!

PR #901, opened 2020-08-08
Fixes for development branch

ashwinkjoseph:
## Description
- Fixes the lock file and maintains old versions while installing the new packages.
- Refactors the tests in functions to use `.runtimeconfig.json`instead of env variables. Also introduces commands in GitHub Actions to set up .`runtimeconfig.json` file at the time of running the tests.
- Added step in action for `Run CI Checks` to run tests for all projects in the monorepo
- Modifies the step in the workflow `Run CI Checks` for running tests on `functions` to create and populate `.runtimeconfig.json` with the config currently deployed to the dev firebase project.
- Adds a check in the step for running tests on `functions` in the `Run CI Checks` workflow to exit with status 0 if secrets aren't passed to it (This is the case when PRs are made from forks)
- Removes override of `setupTests` from `package.json` in web-client and instead makes use of `src/setupTests`.ts which `CRA` references by default
## Motivation
- A conflict in yarn.lock had caused issues with yarn install throwing a 401 and version changes in eslint causing lint errors that didn't exist previously. Refreshing the lock file caused issues with building the site project using gatsby so that wasn't an option
- Earlier Refactor for tests caused issues with running tests in CI as environment variables weren't being used for this and also we would like to use local firebase emulators instead of an actual firebase project for testing purposes.
- `CRA` doesn't support overriding `setupTests` in `package.json` which caused issues running the tests with environment variable `CI=true` but we needed to override `setupTests` as we needed to define `window.matchMedia` and also mock `index DB`
- We exit with status as 0 in the step for running tests in the `Run CI Checks` workflow as not a lot of people would be working on the `functions` project and especially not the new devs. As the secrets aren't passed to workflows executing for PRs from forks of the project, this step will fail for all the PRs made from forks (which is most of PRs from new devs) and will introduce confusion for these new devs. So exiting with status as 0 will ensure that this is avoided. However the tests donot exit with status as 0 when the secrets are present and will execute the actual tests so this will always be executed for PRs within the project (As such we should try to have all PRs with changes in `functions` project to be made from within the repo itself and not from forks
## Release Checklist

- [x] This code has unit tests

## Security Checklist


web-client/package.json
```diff
@@ -59,7 +59,8 @@
     "start": "cross-env SKIP_PREFLIGHT_CHECK=true HTTPS=true react-app-rewired start",
     "clean": "rm -rf build",
     "build": "cross-env react-app-rewired build",
-    "test": "react-app-rewired test --watchAll --env=jest-environment-jsdom-sixteen",
+    "test": "react-app-rewired test --env=jest-environment-jsdom-sixteen",
+    "test:ci": "cross-env CI=true react-scripts test --env=jest-environment-jsdom-sixteen",
```
skspade, line 63: Just curious, does it it work if you do not call `react-scripts` here but `react-app-rewired`?
For example: ` "test:ci": "cross-env CI=true react-app-rewired test --env=jest-environment-jsdom-sixteen"`
IF it's working as is, don't change it. I just think the override issues is caused by using vanilla react-scripts here.

ashwinkjoseph: oh, yea. I might have overlooked this. let me try this out

ashwinkjoseph: yup, it seems to do the trick. I will push it now. thanks 👍

Issue #914, opened 2020-08-13
1.4 As an authenticated user, I want to be able to delete my account


Apfx1945:
As an authenticated user, I want to be able to delete my account.

Acceptance Criteria:

- All the ongoing and pending requests have to be canceled. 

- All the ongoing and pending offers have to be canceled.

- The request attached to the canceled offer has to transition to a canceled state if the offer was accepted for the request. The request will not be affected if the offer wasn’t accepted. 
The user’s personal information has to be replaced with the words “Deleted User” for Display Name,  “deleteduser” for username,  “deleted address” for streetAddress, LatLng object pointing to (0.0) for latLng in all the requests and offers which stored the users' information as snapshots.

- All the user's information is erased from the database by being replaced with the words “Deleted User” for Display Name, or  “deleteduser” for username, or  “deleted address” for all text fields in address objects, LatLng object pointing to (0.0) for latLng. 

--------
### [ZenHub Information](https://app.zenhub.com/workspaces/reach4help-5e8dcbfb14ac087f410cbabb/issues/reach4help/reach4help/914)

*This information is updated automatically. To modify it, please use ZenHub.*

**Belonging to Epics:**

* **[EPIC]** [#910 - Authentication and Profile](https://github.com/reach4help/reach4help/issues/910)

schowdhury8:
@edwardmercado expressed interest in working on this!

divyeshairmeet:
![image](https://user-images.githubusercontent.com/58542990/93378275-d797c300-f879-11ea-9b7a-526b5622dfdd.png)

Made these CSS changes, I'm facing some issues while committing.

Tapudp:
![image](https://user-images.githubusercontent.com/8810813/93378529-378e6980-f87a-11ea-973e-bb64f86c1ef2.png)

Updated these CSS changes, I am facing some husky related errors while committing.

```
husky > commit-msg (node v10.16.3)
⧗   input: css changes for delete button and top-panel
✖   subject may not be empty [subject-empty]
✖   type may not be empty [type-empty]

✖   found 2 problems, 0 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint

husky > commit-msg hook failed (add --no-verify to bypass)
error Command failed with exit code 1.
```

PR #986, opened 2020-09-19
Fix/issue#914

Tapudp:
## Description
This is regarding #914 involves CSS change on settings page.

- Top panel Settings Name and back button
- The list of actions are now absolute position to achieve on top effect on the top panel.
- margin changes for delete account button.

## Release Checklist

## Security Checklist


ashwinkjoseph:
I will resolve this merge conflict

web-client/src/modules/settings/components/SettingsList/SettingsList.tsx
```diff
@@ -69,8 +70,8 @@ const SettingsList: React.FC<SettingsProps> = ({
             </SettingsListCollapsePanel>
           </Collapse>
         </Col>
-      </Row>
-      <Row gutter={[0, 12]}>
+      </SettingsListItemWrapper>
+      <SettingsListItemWrapper topSpace="40px">
```
ashwinkjoseph, line 74: Could you have this stack up one on top of each other using padding and flexbox rather than with top css?
This isn't a very scalable solution for the UI

ashwinkjoseph (changes requested):
@Tapudp I resolved the merge conflict, however, I noticed some critical CSS issues in desktop. The navbar and so on seems to be going out of the page
I haven't committed or pushed this as it might have been introduced during merge, although that is highly unlikely. 
Could you review this issue before merging with development and push an update? then I will resolve the merge conflict
image

web-client/src/modules/settings/components/SettingsList/SettingsList.tsx
Tapudp: resolved this in latest commit.

ashwinkjoseph: Hey @Tapudp , great job on fixing the main CSS issue. It works flawlessly on different screens. 
However, this last one about stacking up one on top of another is still not done.

web-client/src/components/figma/BlockStyles.tsx
```diff
@@ -95,6 +96,22 @@ const SettingsListWrapper = styled.div`
   padding: 10px;
   background-color: ${COLORS.greyBackground};
   height: 100vh;
+  display: flex;
+  flex-direction: column;
+  flex-wrap: wrap;
+`;
+
+const SettingsListItemWrapper = styled.div`
+  margin: -25px 0px;
```
ashwinkjoseph, line 105: Hey @Tapudp , these are hardcoded margin values which means that when I add another item to the list, it breaks and appears on top of the previous items.

Originally this was named as `SettingsListNameWrapper ` and I assume you were going to create another wrapper for another new item, but this isn't the approach to be taken for such UI Components. 

I see that you have used flexbox in the `SettingsListWrapper` and that is the correct direction to proceed however, you need to make use of that here for each item so that we'd have to just use one item wrapper CSS and the items would automatically get stacked one under the other like I asked for in the previous review

web-client/src/modules/settings/components/SettingsList/SettingsList.tsx
```diff
@@ -118,8 +120,8 @@ const SettingsList: React.FC<SettingsProps> = ({
             <span>{t('settings.deleteAccount')}</span>
           </SettingsListButton>
         </Col>
-      </Row>
-      <Row gutter={[0, 12]}>
+      </SettingsListPopupWrapper>
+      <SettingsListItemWrapper>
```
ashwinkjoseph, line 124: You see here, I needed to add another item here for changing address which behaves similarly to how the change name component works.
Having another CSS just for address with hardcoded values for margin here wouldn't be the right approach, instead it should work automatically.

web-client/src/components/figma/BlockStyles.tsx
```diff
@@ -95,6 +96,22 @@ const SettingsListWrapper = styled.div`
...
+`;
+
+const SettingsListPopupWrapper = styled.div`
```
ashwinkjoseph, line 108: What I mentioned above for the Item Wrapper applies here too.
I also see here that you used media queries only for adjusting the margin, if this is for the pop-up, then it's good to have it here like this in a separate CSS but generalize it for all popups.
But if this was intended to increase margin so that the delete option appears under the change name option in the list, please remove it and use flex

ashwinkjoseph (changes requested):
Avoid hardcoding margin pixel values for each item, make it generic using flexbox so that we only have to use an `ItemWrapper` and not separate wrappers like `NameWrapper`, `AddressWrapper` and so on, it's not a scalable solution.

web-client/src/components/figma/BlockStyles.tsx
divyeshairmeet: Working on these chnages.

divyeshairmeet:
I have added another contianer in the `SettingListWrapper` so to support generic `ListItemWrapper`. Though this approach involves both flex-box as well as position absolute for the `SettingsListContainer`, I guess this would never hinder much going forward as we add more list items inside this said container.

ashwinkjoseph (approved):
LGTM!

schowdhury8:
A bit overdue but @all-contributors please add @Tapudp for code

Tapudp:
> A bit overdue but @all-contributors please add @Tapudp for code

Thank you!

PR #997, opened 2020-10-07
Feat 6.1.2 requestflow v1

ethanstrominger:
## Release Checklist

## Security Checklist


ashwinkjoseph:
@ethanstrominger could you fix the errors thrown from the tests here?

web-client/src/modules/requests/constants.tsx
```diff
@@ -7,11 +7,16 @@ import * as Yup from 'yup';
...
+// ?? todo remove Accepted and Open
+export const offersPostType = 'offers';
+export const offersLocationStr = '/requests/'+offersPostType;
+export const requestsPostType = 'requests';
+export const requestsLocationStr = '/requests/'+requestsPostType;
+
 export const AcceptedRequestsLocation = new Location('/requests/accepted');
 export const FindRequestsLocation = new Location('/requests/find');
-export const OpenRequestsLocation = new Location('/requests/open');
+export const OpenRequestsLocation = new Location('/requests/opxxxen');
```
ashwinkjoseph, line 18: There's a typo here and also this route is no longer useful to us. Could you remove it?

web-client/src/modules/requests/constants.tsx
```diff
@@ -7,11 +7,16 @@ import * as Yup from 'yup';
 export const PostListLocation = new Location('/requests/list/:type', {
   type: Yup.string(),
 });
+// ?? todo remove Accepted and Open
+export const offersPostType = 'offers';
+export const offersLocationStr = '/requests/'+offersPostType;
```
ashwinkjoseph, line 12: 1. I thought the plan was to generalize and call it a `Post`. I understand your reasoning for not showing the word post to the end-users but I am not sure if this is the solution for it as in my opinion showing `requests` and then `offers` along with it would be even more confusing than showing `posts`.
2. If showing posts isn't an option, how about changing it to `list`?
3. Why have you chosen to go with this approach of constructing a string rather than using the location package? I don't think this is a good idea and is downgrading the quality of the code unless you have reasoning for it in which case could you explain that here?
4. Why have you chosen to remove and enum and instead use a variable? is there any advantage to doing it in this manner?

web-client/src/modules/requests/containers/RequestsContainer.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import PostsContainer from './PostsContainer';
+import { requestsPostType} from '../constants';
+
+const RequestsContainer: React.FC = () => (
```
ashwinkjoseph, line 5: This container doesn't seem to have any logic in it, so why is it considered as a container?

web-client/src/modules/requests/containers/PostsContainer.tsx
```diff
@@ -129,7 +129,6 @@ const OfferAndRequestPostsContainer: React.FC<PostsProps> = ({
 };
 
 interface PostsProps {
-  postMode: ApplicationPreference; // offer or request
-  status: string;
+  postType: string; // offer or request
```
ashwinkjoseph, line 132: Since the container expects the value of this prop to be one of two things, it's better to have it as an enum so as to leverage typescript features and to avoid confusion for developers

ashwinkjoseph (changes requested):
Hey @ethanstrominger , I have asked a few questions down in the review, could you give me more clarity regarding these questions and also make the changes wherever necessary?

I also see that there are a few linting errors, could you fix those as well?

web-client/src/modules/requests/constants.tsx
schowdhury8: A bit confused here, what's a Post? I'd agree with keeping just requests and offers, at least for the end-user (in the interface) cuz introducing new data structures might just be more confusing. The end-goal is: both a request and offer can be standalone waiting for someone to accept it AND both can be responses to existing requests and offers.

ashwinkjoseph (approved):
LGTM!
I do have a concern about the change in `web-client/src/modules/requests/components/Header.tsx` for removing the code for showing the warning. But as of now, we aren't showing it anywhere in the new design so it should be fine.

ashwinkjoseph:
@ethanstrominger , there seems to be an issue with your code not following prettier configuration for the `MasterPage.tsx`. could you make a commit fixing it?

ashwinkjoseph (approved):
LGTM!

PR #999, opened 2020-10-09
Progress on the new landing page

leadersheir:
## Description
Added **non-functional** -
* navbar/header
* top-section
* middle-section
* footer.

And declared styled components for the sections that still remain to be added.

# Screenshots

Screenshot from 2020-10-09 19-28-13
Screenshot from 2020-10-09 19-28-21
Screenshot from 2020-10-09 19-28-26

## Release Checklist

## Security Checklist


ashwinkjoseph (changes requested):
Hey, 
You shouldn't have an `elements.tsx` file in the `pages` directory. The reasoning is that the `pages` are supposed to only contain files that act as pages in the module.

All components should be in the `components` directory and here these elements are sort of dumb components.

What you could do is to create a file named figma inside components, this way you are conveying that the file contains all the basic components of the module that are styled as per the figma designs.

So `pages/elements` should be moved to `components/figma`.

Another point I noticed is that you had an `elements` file in the `components/Header` directory. This is very subjective but I recommend you do this only if you are sure that these components are not going to be used by any other page in the module. 

Also, another thing to consider while doing this is that unless necessary (maybe there are a lot of UI elements), try to have these in the same file as the file in which they are being used, following the [naming and declaring conventions for the project](https://github.com/reach4help/reach4help/blob/development/web-client/docs/CodingConventions/NAMING_AND_DECLARATION_CONVENTIONS.md). However, I leave this up to your judgement. Also, I recommend changing the name of the file (if you choose to keep the file) to `figma` rather than `elements`

Apart from the above review, the rest of the code looks good to me and ready to be approved

leadersheir:
@ashwinkjoseph I was also thinking about getting rid of the other `elements.tsx` file and merging them into one, creating sections in it with comments that specify which type of styled components should be in that part of the code. As of now, I think one `figma` containing all UI styled components shouldn't be hard to manage. We can break it up in the future if we need to.

And regarding the last part, declaring the UI components in the same file that contains the jsx clutters up the code too much, I think it'll be easier for navigation if they're all kept in a separate file containing only styled components.

I was wondering if there were some sort of naming conventions we were following, so thanks for the link to it, I'll have a look at it and make changes if needed. I'll get onto making the changes you suggested as soon as I'm able to.

leadersheir:
I've made the changes you suggested @ashwinkjoseph :+1:

PR #1002, opened 2020-10-11
Fix routing and update imports

alexjball:
This fixes form issues and an issue where the landing page module was gobbling up routes, showing a blank page.

### DIFF
--- a/web-client/src/modules/landing-page/constants.tsx
+++ b/web-client/src/modules/landing-page/constants.tsx
@@ -1,4 +1,6 @@
 import Location from 'react-app-location';
 
-export const HomePageLocation = new Location('/');
-export const AboutPageLocation = new Location('/about');
+export const urlPrefix = '/home';
+
+export const HomePageLocation = new Location(urlPrefix);
+export const AboutPageLocation = new Location(`${urlPrefix}/about`);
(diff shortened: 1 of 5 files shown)


web-client/src/modules/landing-page/constants.tsx
```diff
@@ -1,4 +1,6 @@
 import Location from 'react-app-location';
 
-export const HomePageLocation = new Location('/');
-export const AboutPageLocation = new Location('/about');
+export const urlPrefix = '/home';
```
leadersheir, line 3: Why change the base url from `'/'` to `'/home'`?

ethanstrominger (approved):

leadersheir (approved):
Looks good. Has no issues with the new landing page. I'm not sure if it's necessary to use `'/home'` as the base route. If it is, that's fine. If it's not, perhaps having `'/'` is the base route is more preferable.

alexjball:
Having a base route is important here. The module base paths are passed to a `Route` without `exact` set, so a base route of "/" matches _all_ URL's. And since the module itself doesn't respond to e.g. /list/find, it doesn't render anything, which is why the screen is white.

We could refactor things to use / and /about, but this was the quickest fix we found. If we refactor eventually, it might be good to make the top-level routes more static (defined centrally).

schowdhury8:
Thanks so much Alex! @all-contributors please add @alexjball for code

PR #1020, opened 2020-10-19
refactor: utils.tsx

ethanstrominger:
Co-authored-by:  @Tapudp <divyeshpar2496@gmail.com>

Refactored utils.tsx to make it more obvious that there may be no addresses and what will be done if there are.  Paired with Divyesh on this and both of us found the new version easier to read.


ashwinkjoseph (approved):
LGTM!

PR #1031, opened 2020-11-05
Bug 1021 tab logic v2

ethanstrominger:
I would like to do more refactoring of this, including:

- see if any duplicated code in OfferPostsContainer and ReqestPostsContainer can be combined without sacrificing readability of either
- add tests

If you don't see anything significantly wrong and you think this improves the code since it fixes a bug, can you merge? I can make changes you want.


ashwinkjoseph (commented):
I notice the need for changing the names at a few places, I will add a more detailed review of this tomorrow

web-client/src/ducks/requests/functions.ts
```diff
@@ -51,39 +51,39 @@ export const setUserRequest = async ({
     .withConverter(RequestFirestoreConverter)
     .set(requestPayload);
 
-export const getOpenRequest = async ({ lat, lng }: IgetOpenRequests) =>
+export const getOpenPost = async ({ lat, lng }: IgetRequestPosts) =>
   functions.httpsCallable('https-api-requests-getRequests')({
```
ashwinkjoseph, line 55: can you make use of firebase directly here to get the data rather than the cloud function? that is what we are aiming for, so we shouldn't use the cloud function here.

web-client/src/ducks/requests/actions.ts
```diff
@@ -3,90 +3,40 @@ import { IRequest, Request } from 'src/models/requests';
...
 
-export const getOpenRequests = (payload: IgetOpenRequests) => (
+export const getRequestPosts = (payload: IgetRequestPosts) => (
   dispatch: Function,
 ) =>
   dispatch({
-    type: GET_OPEN,
-    firebase: getOpenRequestFunc,
+    type: GET_REQUEST_POST,
+    firebase: getOpenPostFunc,
```
ashwinkjoseph, line 27: This will only give you the posts that have status as open. We need posts of all statuses here. I have already mentioned as separate comment in the functions file for using firebase directly.

web-client/src/ducks/requests/actions.ts
```diff
@@ -3,90 +3,40 @@ import { IRequest, Request } from 'src/models/requests';
...
 
-export const getAcceptedRequests = (payload: IgetOpenRequests) => (
+export const getOfferPosts = (payload: IgetRequestPosts) => (
   dispatch: Function,
 ) =>
   dispatch({
-    type: GET_ACCEPTED,
-    firebase: getAcceptedRequestFunc,
+    type: GET_OFFER_POST,
+    firebase: getAcceptedPostFunc,
```
ashwinkjoseph, line 36: This would only give you posts which you created which have been accepted. This function is supposed to give you the posts for which you responded as well.
I have already mentioned as a separate comment on the functions page for using firebase directly instead of the cloud functions

ashwinkjoseph (changes requested):
Can you switch to using firebase directly here?
There are a few changes to variable names that are required as well, I will put details regarding those tomorrow but they aren't super important, I can make them as a separate PR after merging this if you would like to proceed in that manner or we can wait till tomorrow and make the changes in this PR itself.

However, the functions should make use of firebase directly and should no longer make use of the cloud functions

web-client/src/ducks/requests/functions.ts
ethanstrominger: Yes, I am working on that, but would like it to be a separate pull request.

web-client/src/ducks/requests/actions.ts
ethanstrominger: I know, I want to do this as part of a separate pull request.  What is in develop branch has same issue.

web-client/src/ducks/requests/actions.ts
ethanstrominger: I know, I want to do this as part of a separate pull request. What is in develop branch has same issue.

ashwinkjoseph (approved):
Approving with the knowledge of changing in another PR

PR #1041, opened 2020-11-14
[SHAYAN - REVIEW YOUR COMMENTS AND DECIDE WHAT TO DO] feat: reac-69 and reac-70. Implemented the rest of the about page

14ssharp:
## Description

Finished up working on the about page. The pictures are all just from the github contributors, so I am sure there will be a lot of changes still needed there. I have made structured this in a way that shouldn't be too hard to edit. The image sources are in contributors.tsx, which is imported into AboutPage.tsx. You can change a picture by editing one of those image sources.

## Motivation

Reac-69 and Reac-70

## Testing Guidelines

1) Run the development server
2) Visually confirm the webpage layout

## Release Checklist

## Security Checklist

## Additional Notes
image
image

### DIFF
--- a/web-client/src/components/MenuDrawer/MenuDrawer.tsx
+++ b/web-client/src/components/MenuDrawer/MenuDrawer.tsx
@@ -6,7 +6,7 @@ import styled from 'styled-components';
 
 import { InformationModal, makeLocalStorageKey } from '../Modals/OneTimeModal';
 import SideBottomMenu from '../SideBottomMenu/SideBottomMenu';
-import SideDrawerProfile from '../SideDrawerProfile/SideDrawerProfile';
+// import SideDrawerProfile from '../SideDrawerProfile/SideDrawerProfile';
 import SideTopMenu from '../SideTopMenu/SideTopMenu';
 
 const MenuDrawer: React.FC<IMenuDrawerProps> = ({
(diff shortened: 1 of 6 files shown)


ashwinkjoseph (changes requested):
Hey, the PR looks good to me, but could you retrieve the user list from https://github.com/reach4help/reach4help/blob/development/.all-contributorsrc instead of keeping it as a static file?

I have requested this as a change in this review, but if you want I can merge this in and accept the change as a new PR. If this is what you want, let me know by commenting here and mentioning me or by messaging me on slack

schowdhury8:
On that note, @all-contributors please add @14ssharp for code

schowdhury8:
Actually, on second thought @ashwinkjoseph , all-contributors doesn't usually represent the designers, marketers or product owners. So I think a separate file isn't a bad idea, as it'll give us a bit more flexibility to convey other information like titles and whatnot. 
@14ssharp would you be able to make your list of people similar to something like this instead: https://github.com/reach4help/reach4help/blob/526636ba15e0e55648652ee389341c84e1cabb9f/site/src/sections/team/index.js#L9
It'll also save you time from trying to hardcode each person cuz you can just map through 'em.

ashwinkjoseph:
> Actually, on second thought @ashwinkjoseph , all-contributors doesn't usually represent the designers, marketers or product owners. So I think a separate file isn't a bad idea, as it'll give us a bit more flexibility to convey other information like titles and whatnot.
> @14ssharp would you be able to make your list of people similar to something like this instead:
> 
> https://github.com/reach4help/reach4help/blob/526636ba15e0e55648652ee389341c84e1cabb9f/site/src/sections/team/index.js#L9-L30
> 
> 
> It'll also save you time from trying to hardcode each person cuz you can just map through 'em like this:
> https://github.com/reach4help/reach4help/blob/526636ba15e0e55648652ee389341c84e1cabb9f/site/src/sections/team/index.js#L63-L68

Yea, that sounds good too, if we are okay with managing that list separately, but it shouldn't be too hard considering it's JSON now

PR #1073, opened 2020-12-12
Feature/create post

mcverter:
Create Post Rough Draft.
Implements basic functionality of CreatePost.  Needs styling to resemble:
https://jira.reach4help.org/browse/REAC-9


web-client/src/modules/create/components/NewAddressModal.tsx
```diff
@@ -0,0 +1,158 @@
...
+            privilegedInformation.addresses = addressesToSave;
+            dispatch(
+              updateUserPrivilegedInformation(userId, privilegedInformation),
+            );
+          }
+        }
+      })
+      .catch(error => {
+        // eslint-disable-next-line no-console
+        console.error('could not geocode', error);
```
alexjball, line 74: Should this failure be passed back to the container? I think currently if this fails, the modal will just remain visible. passing the address to teh container would allow you to handle this case there.

web-client/src/modules/create/components/PostDetails.tsx
```diff
@@ -0,0 +1,170 @@
...
+                  message: t('newRequest.form.type_error_message'),
+                },
+              ]}
+            >
+              <Input />
+            </Form.Item>
+
+            <Form.Item
+              label={t('newRequest.form.type')}
+              name="type"
```
alexjball: Just to confirm, can this be any value or should it be a dropdown? behavior on development is a dropdown I believe

web-client/src/modules/create/components/PostMap.tsx
```diff
@@ -0,0 +1,134 @@
+import { Button, Select } from 'antd';
+import React, { useState } from 'react';
+import TitleWithOrangeUnderline from 'src/components/TitleWithOrangeUnderline/TitleWithOrangeUnderline';
+import { IUserAddress } from 'src/models/users/privilegedInformation';
+import { NewAddressModal } from 'src/modules/create/components/NewAddressModal';
+import { COLORS } from 'src/theme/colors';
+import styled from 'styled-components';
+
+import WebClientMap from '../../../components/WebClientMap/WebClientMap';
```
alexjball: nit: use an absolute 'src...' path here and in other files for consistency

web-client/src/modules/create/components/PostMap.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const [showAddressModal, setShowAddressModal] = useState<boolean>(false);
+  const modalSuccess = value => {
+    setPostLocation(value);
+    setShowAddressModal(false);
+  };
+  const closeModal = () => {
+    setShowAddressModal(false);
+  };
+
+  const handleNameChange = value => {
```
alexjball: nit: consider renaming to handleAddressChange as it is more specific

web-client/src/modules/create/components/PostSummary.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const history = useHistory();
+
+  return (
+    <>
+      {showConfirmationPage && (
+        <PostConfirmation
+          showModal={showConfirmationPage}
+          closeModal={() => {
+            setShowConfirmationPage(false);
+            // because I could observe race conditions in cloud function
```
alexjball: does this mean that the my request posts page loads before the request is submitted to firebase, and so does not show the new request? I believe you could handle this by moving the history.replace call into the container, and waiting for the request submission to complete before calling it.

web-client/src/modules/create/containers/CreatePostContainer.tsx
```diff
@@ -1,336 +1,152 @@
...
+      ),
+    },
+    {
+      title: 'Summary',
+      component: (
+        <PostSummary
+          prevHandler={moveBackwards}
+          postDetails={postDetails}
+          postLocation={postLocation}
+          submitRequest={submitPost}
```
alexjball: If nit: rename this prop to submitPost since it sounds like this will support requests and offers

web-client/src/modules/create/components/NewAddressModal.tsx
```diff
@@ -0,0 +1,158 @@
...
+          </Row>
+          {shouldSave && (
+            <Row gutter={[16, 6]}>
+              <Col span={24}>
+                <Form.Item
+                  name="name"
+                  required
+                  label={t('settings.changeAddressForm.nameOfAddress')}
+                >
+                  <Input id="nameofAddress" />
```
alexjball, line 141: Can you validate that this does not conflict with an existing address name?

web-client/src/modules/create/components/NewAddressModal.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+  const addresses = useSelector(
+    (state: AppState) => state.profile.privilegedInformation?.addresses,
+  );
+  const privilegedInformation = useSelector(
+    (state: AppState) => state.profile.privilegedInformation,
+  );
+  const userId = useSelector((state: AppState) => state.auth.user?.uid);
+
+  const handleFinish = () => {
```
alexjball, line 33: Components generally should not interact with redux or call services like geocode directly--that is the job of the container. This separation makes it possible in principle to render components in isolation, without needing to mock the store or set up application state. Consider passing the address object in modalSuccess, and bubbling it back to the container, where it calls geocode and dispatches the update

alexjball (dismissed):
Looks and works great! As discussed offline, I see duplicate requests in my requests after I submit a new request.

ashwinkjoseph:
Hey @mcverter if you have made the changes that alex requested, could you either fix the broken test or could you disable that test for the time being? the test is looking for parameters from the older version of create

web-client/src/modules/create/components/NewAddressModal.tsx
mcverter: This wrapped component interacts directly with the geocoder.   The CreatePostContainer does not need to know about it.

web-client/src/modules/create/components/NewAddressModal.tsx
mcverter: Fixed. onGeocodeFail() defined in CreatePostContainer

web-client/src/modules/create/components/NewAddressModal.tsx
mcverter: just added some code in handleFinish()

web-client/src/modules/create/components/PostDetails.tsx
mcverter: Fixed to Dropdown

web-client/src/modules/create/components/PostMap.tsx
mcverter: Done

mcverter:
@alexjball @ashwinkjoseph  Please review and comment or merge.  It still needs some design but it works okay and addressses Alex's comments from the last week.  

There is an error in the CI check which I do not think is related to my code:
chunk 2 [mini-css-extract-plugin]
Conflicting order. Following module has been added:
 * css ../node_modules/css-loader/dist/cjs.js??ref--6-oneOf-7-1!../node_modules/postcss-loader/src??postcss!../node_modules/less-loader/dist/cjs.js??ref--6-oneOf-7-3!../node_modules/antd/es/checkbox/style/index.less
despite it was not able to fulfill desired ordering with these modules:
 * css ../node_modules/css-loader/dist/cjs.js??ref--6-oneOf-7-1!../node_modules/postcss-loader/src??postcss!../node_modules/less-loader/dist/cjs.js??ref--6-oneOf-7-3!../node_modules/antd/es/collapse/style/index.less
   - couldn't fulfill desired order of chunk group(s) ,

alexjball (approved):
Thanks Mitchell!

PR #1079, opened 2020-12-20
Create post bugfix

mcverter:
Fixes merging of "create-post" branch.  
https://github.com/reach4help/reach4help/commit/a346879d3521849dffe7dee4d4941320388c53cd
Putting an ANTD Checkbox within a ANTD Modal produces an error in the css file


web-client/src/modules/create/containers/CreatePostContainer.tsx
```diff
@@ -1,336 +1,196 @@
...
+  const [stepNumber, setStepNumber] = useState(0);
+  const moveForwards = () => setStepNumber(stepNumber + 1);
+  const moveBackwards = () => setStepNumber(stepNumber - 1);
+
+  /* PostDetails */
+  const getTypes = () => [t('modules.create.defaults.postDetails.type')];
+  const [postDetails, setPostDetails] = useState<IPostDetails>({
+    title: t('modules.create.defaults.postDetails.title'),
+    body: t('modules.create.defaults.postDetails.body'),
+    type: getTypes()[0],
```
shayantist: Looks great to me overall functionality wise, but hardcoding default values might cause people to miss filling them out themselves. Instead probably best to keep these as an example `placeholder` for each `Input` component in `PostDetailsStep.tsx`

web-client/src/modules/create/components/PostDetailsStep.tsx
```diff
@@ -0,0 +1,179 @@
...
+            label={t('modules.create.postDetails.titleLabel')}
+            name="title"
+            rules={[
+              {
+                required: true,
+                message: t('newRequest.form.title_error_message'),
+              },
+            ]}
+          >
+            <Input />
```
shayantist: In reference to my other comment, maybe replace this with: `<Input placeholder="[default/example value]"/>`

shayantist (changes requested):

web-client/src/modules/create/containers/CreatePostContainer.tsx
mcverter: fixed

web-client/src/modules/create/components/PostDetailsStep.tsx
mcverter: fixed

shayantist (approved):

PR #1085, opened 2020-12-26
Aboutpage2

Danieleclima:
I got the pictures to appear I still need to modify the team titles so that they appear on top.


ashwinkjoseph (changes requested):
Hey, can you add the .env.EXAMPLE file back in? I noticed two empty files as well, can you look into whether they can be removed as well?

shayantist:
This is a great start, Daniel! Looking forward to seeing the end result. 

> Hey, can you add the .env.EXAMPLE file back in? I noticed two empty files as well, can you look into whether they can be removed as well?

I've resolved this. @Danieleclima when would you be able to finish the team titles? 
Also with this PR, does #1081 need to be open?

Danieleclima:
> This is a great start, Daniel! Looking forward to seeing the end result.
> 
> > Thanks! sorry for the radio silence, been sick the last couple days.
I tried to post the final changes but it looks like somebody has made changes and added media queries to the styled components file, is that posible?

PR #1095, opened 2021-01-04
Triggers for creating a post or response

ashwinkjoseph:
## Description
This PR adds triggers that are responsible for validating and performing side effects of creating the post or response created and also for adding the posts and responses into algolia.

## Motivation
We needed to populate the new algolia indices with posts and responses created with the new data model

## Release Checklist

## Security Checklist

## Additional Notes

The tests haven't yet been updated, we need to update the tests so that it no longer tests for data in algolia in the request and offer test suits and instead does so in the post test suite, we also need a test suite for posts


functions/src/offers/functions.ts
```diff
@@ -7,7 +7,6 @@ import { IOffer, Offer, OfferStatus } from '../models/offers';
 import { IRequest, Request, RequestStatus } from '../models/requests';
```
ethanstrominger, line 7: Why are you making changes to offers?  We are not using offers for Find and "offers" by itself is no longer a concept.

web-client/src/modules/MyPosts/containers/FindRequestsAlgoliaContainer.tsx
```diff
@@ -32,7 +32,11 @@ const FindRequestsContainer: React.FC = () => {
   );
```
ethanstrominger, line 32: Nothing in here is specific to finding requests, so it should be put in a more general module.

ethanstrominger (dismissed):

functions/src/offers/functions.ts
ashwinkjoseph: I removed the functions I had written previously for algolia to index offers and requests, so I had to update the previously existing triggers for offers and requests.
I could just delete these triggers as well but not all of the app is still migrated and the version deployed on app.dev.reach4help.org is still dependant on these triggers

web-client/src/modules/MyPosts/containers/FindRequestsAlgoliaContainer.tsx
ashwinkjoseph: I don't understand how you came to that conclusion, why do you think nothing here is specific to finding requests?

ethanstrominger: Would this apply to requests and offers?  If yes, change Request to Post for filename and code.  If no, what is it that makes it specific to requests?

functions/src/offers/functions.ts
ethanstrominger: You are trying to support two versions at once?  That could get tricky, do we really need to do that?  When the new app is implemented, do we need the non-Algolia cloud functions at all?

web-client/src/modules/MyPosts/containers/FindRequestsAlgoliaContainer.tsx
ashwinkjoseph: ah yes, this was made about a month ago when we were still deciding these changes, will make the change

functions/src/offers/functions.ts
ashwinkjoseph: no that wouldn't be tricky, the original idea was to go with a phased-out refactor. If the version on dev isn't a problem, then this can be removed

PR #1098, opened 2021-01-08
[SHAYAN: PLEASE REVIEW] Changes to about page

Danieleclima:
I've changed the team titles and the background color for each of the teams.


shayantist (changes requested):
Just to reiterate the changes I mentioned on Slack so that it doesn't get lost there: 
- Could you make each category the full width of the screen?
- Instead of scrolling horizontally, just make it wrap vertically (flex-wrap) for each category (much better UX that way)
- Under each profile picture, add a text box for a one/two word description of the person (will be pulled from the JSON)

PR #1105, opened 2021-01-19
makes more space by removing title elements.  submitPost() occurs aft…

mcverter:
In order to create more space for the display elements, I have removed the Titles for each screen.  They are already given in the step-header.

submitPost() now occurs after the user presses the submit button.  Before it occured after closing the confirmation modal

### DIFF
--- a/web-client/src/modules/create/components/PostSummaryStep.tsx
+++ b/web-client/src/modules/create/components/PostSummaryStep.tsx
@@ -2,7 +2,6 @@ import { ArrowLeftOutlined, StarOutlined } from '@ant-design/icons';
 import React, { useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useHistory } from 'react-router-dom';
-import TitleWithUnderline from 'src/components/TitleWithUnderline/TitleWithUnderline';
 import {
   AddressDisplay,
   ButtonsContainer,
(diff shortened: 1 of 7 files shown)


web-client/src/modules/create/components/PostSummaryStep.tsx
```diff
@@ -28,17 +27,17 @@ const PostSummary: React.FC<PostSummaryProps> = ({
   const [showConfirmationPage, setShowConfirmationPage] = useState(false);
   const history = useHistory();
 
-  const handleSubmit = () => setShowConfirmationPage(true);
+  const handleSubmit = () => {
+    submitPost();
```
alexjball: Should we await this before showing the modal? What should we do if this fails for some reason?

alexjball (changes requested):

web-client/src/modules/create/components/PostSummaryStep.tsx
mcverter: Point well taken.   I thought the delay time for the interaction with the modal would be sufficient for the new request to propagate to the firebase database.  But I will try to take advantage of the promise.

mcverter: @alexjball fixed

alexjball (approved):

PR #1183, opened 2021-02-16
made create new in side menu appaer to non logged users

mcverter:
Mohammed's changes on dev


shayantist:
Looks good to me. That should be visible to direct you to sign in anyway.

PR #1386, opened 2021-06-20
Map algolia

ethanstrominger:
For environmental vars added, see https://docs.google.com/document/d/1DNklTnhFZWYsJKnOzHK3VhFznAck1UP9biVd9P9wyok/edit

Issues
  - This will. not be protected from development but data is not so sensitive that this should be an issue.  We either need to get algolia to upgrade reach4help "application" (equivalent of an account) which will then let us provide fine grained security or get algolia to give reach4help-dev have same privs as reach4help.
  - If the json includes invalid org type or service type the app may crash.  Logged issue #1426 for this.

Change REACT_APP_ALGOLIA_INDEX_NAME to markers-production

To migrate the data:

- Download json file
- Create a .env file with the four algolia variables and save in map/src/algolia-scripts
cd map/src/algolia-scripts
node import-marker-json-to-algolia.mjs ~/Downloads/markers-prod.json markers-production


map/src/data/dataDriver.ts
```diff
@@ -2,34 +2,23 @@ import {
...
-  MARKERS_STORAGE_PATH,
 } from '@reach4help/model/lib/markers';
-import * as firebase from 'firebase/app';
-
-// eslint-disable-next-line import/no-duplicates
-import 'firebase/firestore';
-// eslint-disable-next-line import/no-duplicates
-import 'firebase/storage';
-// eslint-disable-next-line import/no-duplicates
-import 'firebase/analytics';
```
shayantist, line 15: Until we integrate Algolia analytics, we shouldn't remove this.

map/src/data/dataDriver.ts
```diff
@@ -147,22 +138,28 @@ export const removeInformationListener = (l: InformationListener) => {
   listeners.delete(l);
 };
 
+// TODO:
+// X Import marker data into algoia index
+// X Set up Algoia API keys for the index
+// - Replace below code with code to return a promise to load from algolia
+// - look at index settings to set what is returned from Algoli
```
shayantist: Has this been done yet?

map/src/algoliaUtil.js
```diff
@@ -0,0 +1,90 @@
+/* eslint-disable no-param-reassign */
+/* eslint-disable no-console */
+import algoliasearch from 'algoliasearch';
+import dotenv from 'dotenv';
+import { v4 as uuidv4 } from 'uuid';
+
+dotenv.config();
+const algoliaAdminKey = process.env.REACT_APP_ALGOLIA_ADMIN_KEY || 'undefined';
+const algoliaAppId = process.env.REACT_APP_ALGOLIA_APP_ID || 'undefined';
+const client = algoliasearch(algoliaAppId, algoliaAdminKey);
```
shayantist: Why are these being defined again if they're in [dataDriver.ts](https://github.com/reach4help/reach4help/blob/map-algolia/map/src/data/dataDriver.ts)? What purpose does this file serve?

shayantist (changes requested):
Really well done, @ethanstrominger and @jensren . The transition from Firestore to Algolia feels almost seamless functionality-wise, while providing huge performance gains. Once we implement full-text search and migrate the entire Firestore db over to Algolia (bonus points for geo-search), then we'll be ready to merge this. 

What purpose do the files under [algolia_scripts](https://github.com/reach4help/reach4help/tree/map-algolia/map/src/algolia-scripts) serve?

I see a lot of changed files under [parse_scripts](https://github.com/reach4help/reach4help/tree/map-algolia/map/src/parse-scripts), which likely overwrites our data work on the map branch. Since this PR is only for changes related to Algolia, better to remove them to prevent conflicts. Check some of my other comments on specific files as well.

map/src/data/dataDriver.ts
ethanstrominger: This will not affect firebase analytics.    Firebase is no longer used by dataDriver.ts (previously named fireBase.ts)

PR #1406, opened 2021-07-04
Add more test markers, fix some problems, add docs

jensren:
Added another json file with more test markers, including hidden markers, different organization types, and varying services offered.

Removed casting the filter to any in order to dynamically index into the filter with a string by casting filterScreenField as a key of Filter.
image

I could not fully refactor out the translationObject.services.any, since the organization markers do not have a translated label for any yet. Instead, I made translationObject.services.any the default option if translationObject[translationKey]['any'] is unavailable, and otherwise translationObject[translationKey]['any'] is used.
image

Added some javadocs for the more confusing parts of the code.

### DIFF
--- a/map/src/components/drop-down.tsx
+++ b/map/src/components/drop-down.tsx
@@ -38,7 +38,7 @@ class DropDown extends React.Component<Props, {}> {
     translationObject: any,
     propKey: string,
     valueKey: string,
-  ): string => {
+  ): string | undefined => {
     let val = translationObject;
     const keys = propKey.split('.');
     keys.forEach(key => {
@@ -61,17 +61,30 @@ class DropDown extends React.Component<Props, {}> {
         value,
         {
           value,
-          label: t(lang, translationObject =>
-            this.lookUpValue(translationObject, translationKey, value),
+          label: t(
+            lang,
+            translationObject =>
+              this.lookUpValue(
+                translationObject,
+                translationKey,
+                value,
+              ) as string,
           ),
         },
       ]),
     );
 
-    // TODO: refactor .services
     const any: OptionType = {
       value: undefined,
-      label: t(lang, translationObject => translationObject.services.any),
+      label: t(lang, translationObject =>
+        this.lookUpValue(translationObject, translationKey, 'any')
+          ? (this.lookUpValue(
+              translationObject,
+              translationKey,
+              'any',
+            ) as string)
+          : translationObject.services.any,
+      ),
     };
 
     const options: OptionType[] = [any, ...optionsMap.values()];
--- a/map/src/components/drop-down.tsx
+++ b/map/src/components/drop-down.tsx
@@ -76,12 +76,9 @@ class DropDown extends React.Component<Props, {}> {
 
     const options: OptionType[] = [any, ...optionsMap.values()];
 
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const searchInOptions = (filterScreen: any) =>
-      filterScreen[filterScreenField];
     const value =
-      typeof searchInOptions !== 'undefined'
-        ? optionsMap.get(searchInOptions(filter))
+      typeof filter[filterScreenField as keyof Filter] !== 'undefined'
+        ? optionsMap.get(filter[filterScreenField as keyof Filter] as string)
         : undefined;
 
     return (
--- a/map/src/components/drop-down.tsx
+++ b/map/src/components/drop-down.tsx
@@ -12,6 +12,14 @@ type OptionType = {
   label: string;
 };
 
+/**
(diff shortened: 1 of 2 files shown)


map/src/components/drop-down.tsx
```diff
@@ -32,13 +40,22 @@ class DropDown extends React.Component<Props, {}> {
     }
   };
 
+  /**
+   * Returns the translated word at translationObject.propKey.valueKey if it exists, else undefined.
+   * For example, translationObject.services.org returns "Organization" if the translationObject
```
ethanstrominger, line 45: Give code example.  For example:
   lookupValue (translationObject, 'visibility.filter", "visible") 
returns "Visible markers" if the visibility.filter.visible property of translationObject.visibilty.filter is "Visible marker".

map/src/components/drop-down.tsx
```diff
@@ -32,13 +40,22 @@ class DropDown extends React.Component<Props, {}> {
     }
   };
 
+  /**
```
ethanstrominger, line 43: Change above method. name changeService to onChange

map/src/components/drop-down.tsx
```diff
@@ -12,6 +12,14 @@ type OptionType = {
   label: string;
 };
 
+/**
+ * @param className the class name
```
ethanstrominger, line 16: The className of the Select component

map/src/components/drop-down.tsx
```diff
@@ -12,6 +12,14 @@ type OptionType = {
...
+ * @param translationKey a dot-separated list of keys used to index a translationObject generated from json
+ * @param filterScreenField the property of filter this drop-down updates
+ * @param dropDownValues the drop down values
```
ethanstrominger, line 19: the property => name of the property

map/src/components/drop-down.tsx
```diff
@@ -12,6 +12,14 @@ type OptionType = {
...
+ * @param filter the filter applied to the map
```
ethanstrominger, line 20: the filter state that tracks the values entered into the filter dialog

map/src/components/drop-down.tsx
```diff
@@ -12,6 +12,14 @@ type OptionType = {
...
+ * @param updateFilter a callback that updates filter
+ */
 interface Props {
```
ethanstrominger, line 23: Rename to DropDownProps

map/src/components/drop-down.tsx
```diff
@@ -61,27 +78,37 @@ class DropDown extends React.Component<Props, {}> {
         value,
         {
           value,
-          label: t(lang, translationObject =>
-            this.lookUpValue(translationObject, translationKey, value),
+          label: t(
```
ethanstrominger, line 81: Change above select to selectComponent

map/src/components/drop-down.tsx
```diff
@@ -61,27 +78,37 @@ class DropDown extends React.Component<Props, {}> {
         value,
```
ethanstrominger, line 78: change three variables into three functions that take lang as a parameter and move up a level

map/src/components/drop-down.tsx
```diff
@@ -61,27 +78,37 @@ class DropDown extends React.Component<Props, {}> {
...
       value: undefined,
-      label: t(lang, translationObject => translationObject.services.any),
+      label: t(lang, translationObject =>
+        this.lookUpValue(translationObject, translationKey, 'any')
+          ? (this.lookUpValue(
+              translationObject,
+              translationKey,
+              'any',
+            ) as string)
+          : translationObject.services.any,
```
ethanstrominger, line 103: should be translationObject.any so only needs to be translated once, but not critical - we can discuss when you implement multi-select, as "Any" will disappear from two of the drop downs.

map/src/components/drop-down.tsx
```diff
@@ -61,27 +78,37 @@ class DropDown extends React.Component<Props, {}> {
...
 
     const options: OptionType[] = [any, ...optionsMap.values()];
 
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const searchInOptions = (filterScreen: any) =>
-      filterScreen[filterScreenField];
     const value =
-      typeof searchInOptions !== 'undefined'
-        ? optionsMap.get(searchInOptions(filter))
+      typeof filter[filterScreenField as keyof Filter] !== 'undefined'
```
ethanstrominger, line 110: Might be more clear if condition separated into a variable.  I keep finding myself rereading this so I understand it.

filterKey = filterScreenField as keyof Filter;
isTypeDefined = filter[filterKey] !== 'undefined';

Then, use variables below.

Issue #1426, opened 2021-07-18
Handle invalid and null org and service types


ethanstrominger:
If a marker has an invalid org type, the system will give you an error because filters expect types to have certain values.  Supply type should also be checked.

PR #1487, opened 2021-08-02
Map algolia

ethanstrominger:
For environmental vars added, see https://docs.google.com/document/d/1DNklTnhFZWYsJKnOzHK3VhFznAck1UP9biVd9P9wyok/edit

Issues

This will. not be protected from development but data is not so sensitive that this should be an issue. We either need to get algolia to upgrade reach4help "application" (equivalent of an account) which will then let us provide fine grained security or get algolia to give reach4help-dev have same privs as reach4help.
If the json includes invalid org type or service type the app may crash. Logged issue Handle invalid and null org and service types #1426 for this.
Change REACT_APP_ALGOLIA_INDEX_NAME to markers-production

To migrate the data:

Download json file
Create a .env file with the four algolia variables and save in map/src/algolia-scripts
cd map/src/algolia-scripts
node import-marker-json-to-algolia.mjs ~/Downloads/markers-prod.json markers-production


map/src/components/map-layout.tsx
```diff
@@ -22,63 +28,150 @@ interface Props {
...
+                    <span className="grow" />
+                    <Chevron className="toggle chevron" />
+                  </button>
+                  <form onSubmit={this.handleSubmit} className="form">
+                    <div className="row">
+                      <Search className="search" searchInputId="main" />
+                    </div>
+                    <h3 className="filter filter-heading">
+                      {t(lang, s => s.filterForm.title)}
+                    </h3>
```
shayantist, line 102: As discussed, the "Filter Markers" bar isn't necessary to help save us some extra space. 
Also, "Organization Type" and "Services Offered" can also be removed and added to the Text Box placeholder ("Select organization type" and "Select services you need"). 
image

shayantist (commented):
@ethanstrominger have we migrated all our Firebase markers over to an Algolia index (markers-proto in this case)? If not, let's set that as a point for our next meeting.

shayantist:
Looks good to me for now!
Merging to our `staging` branch before moving to `map`/`production` for further UX design and QA

PR #1626, opened 2022-01-02
Staging to prod: improve map performance

ethanstrominger:
Look for notes on "logic change" below - all else was renaming or minor refactoring
 - footer.tsx - renaming
 - map-layout.tsx - renaming
 - deploy-to-firebase-production.yml: brings staging up to par with production
- add-information.tsx: replaces incorrectly spelled attribute with existing attribute
- google-maps.ts: renamed function and changed zoom from 3 to 10
- map-state.ts: renamed vars, got rid of MARKER_SET_KEYS as it was only providing a constant
- map.tsx
  - got rid of MARKER_SET_KEYS
  - renaming
  - removed unused DataSet related to MARKER_SET_KEYS
  - removed unused getMarkerId
  - **logic change: simplified logic in componentDidMount for getting center location. Unduplicated code and changed ".then()" to await and extracted getting center logic to function getCenterLocation**
  - removed MARKER_SET_KEYS loop which only had one constant, kept logic in loop
  - set zoom to 10
  - fixed exception when marker missing info.info.types for filter - found at least one record with that issue in prod
  - renamed informationUpdated to updateClusterAndMarkersListener
    and refactored.
   - renamed informationUpdated to updateClusterAndMarkersListener
    and refactored.
    - Added some jsdoc (javadoc)
- results.tsx
  - **logic change: added conditional checks in render for info.contentBody and info.contact.  Display message try again if both don't exist.**
- dataDriver.tsx
    - renaming
    - **logic change: repurposed hidden/visible to initial/details.  initial is used for getting data within displayed map area with minimum attributes only .  details retrieves all data, does not limit by area, and includes all attributes.**
    - **logic change: modified informationUpdated to use allMarkers which is populated at beginning with data from initialMarkers and detailMarkers**
- removed unused firebase.ts
- added util.tsx debugLog function.  
- yarn.lock version updates
- **logic change: created Wiki for reverse engineering map logic.  See https://github.com/reach4help/reach4help/wiki**


shayantist:
Great cleanup on the logic for getCenterLocation and for adding the conditional checks on data. The performance difference is pretty impressive now; I tried testing it out on Slow 3G and it's very remarkable. The switch over to Algolia's definitely paying off! 

Leaving a few minor comments in the add-information.tsx file but apart from that, this is brilliant! Thank you, Ethan!

shayantist (changes requested):
I realize the "action n" texts on lines 1124 and 1156 were probably for debugging, but might be a good idea to remove for production.

shayantist (approved):

PR #1776, opened 2022-03-31
Add accordion to show organization results detail

pratikmdhr:
## Description
This PR adds accordion to the organization search results and will show the details of the organization upon expanding. Previously, clicking an organization would show the details in a new view. Now it will appear right below the name of the organization.

The result details will still be shown in a new view if the organization markers are directly clicked on the map. 

## Motivation
Improve user experience

## Release Checklist

## Security Checklist

## Additional Notes
**Before clicking**
image

**After clicking**
image

**When the map marker is directly clicked**
image

### DIFF
--- a/map/src/App.tsx
+++ b/map/src/App.tsx
@@ -26,6 +26,7 @@ interface State {
   nextResults: ResultsSet | null;
   resultsOpen: boolean;
   selectedResult: MarkerIdAndInfo | null;
+  showResultDetails: boolean;
   updateResultsCallback: (() => void) | null;
   lang: i18n.Language;
   inFrame: boolean;
@@ -40,6 +41,7 @@ class App extends React.Component<Props, State> {
       results: null,
       nextResults: null,
       resultsOpen: false,
+      showResultDetails: false,
       selectedResult: null,
       updateResultsCallback: null,
       lang: i18n.getLanguage(),
@@ -68,8 +70,14 @@ class App extends React.Component<Props, State> {
     this.setState({ updateResultsCallback: callback });
   };
 
-  private setSelectedResult = (selectedResult: MarkerIdAndInfo | null) => {
-    this.setState({ selectedResult });
+  private setSelectedResult = (
+    selectedResult: MarkerIdAndInfo | null,
+    showDetails = false,
+  ) => {
+    this.setState({
+      selectedResult,
+      showResultDetails: showDetails,
+    });
   };
 
   private setNextResults = (nextResults: ResultsSet | null) => {
@@ -128,6 +136,7 @@ class App extends React.Component<Props, State> {
       results,
       nextResults,
       resultsOpen,
+      showResultDetails,
       selectedResult,
       page,
       lang,
@@ -181,6 +190,7 @@ class App extends React.Component<Props, State> {
                     setSelectedResult={this.setSelectedResult}
                     updateResults={this.updateResults}
                     showMoreResults={this.showMoreResults}
+                    showResultDetails={showResultDetails}
                   />
                 ),
               }}
--- a/map/src/App.tsx
+++ b/map/src/App.tsx
@@ -26,7 +26,7 @@ interface State {
   nextResults: ResultsSet | null;
   resultsOpen: boolean;
   selectedResult: MarkerIdAndInfo | null;
-  showResultDetails: boolean;
+  mapMarkerClicked: boolean;
   updateResultsCallback: (() => void) | null;
   lang: i18n.Language;
   inFrame: boolean;
@@ -41,7 +41,7 @@ class App extends React.Component<Props, State> {
       results: null,
       nextResults: null,
       resultsOpen: false,
-      showResultDetails: false,
+      mapMarkerClicked: false,
       selectedResult: null,
       updateResultsCallback: null,
       lang: i18n.getLanguage(),
@@ -72,11 +72,11 @@ class App extends React.Component<Props, State> {
 
   private setSelectedResult = (
     selectedResult: MarkerIdAndInfo | null,
-    showDetails = false,
+    markerClicked = false,
   ) => {
     this.setState({
       selectedResult,
-      showResultDetails: showDetails,
+      mapMarkerClicked: markerClicked,
     });
   };
 
@@ -136,7 +136,7 @@ class App extends React.Component<Props, State> {
       results,
       nextResults,
       resultsOpen,
-      showResultDetails,
+      mapMarkerClicked,
       selectedResult,
       page,
       lang,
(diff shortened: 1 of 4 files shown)


map/src/components/map.tsx
```diff
@@ -78,7 +78,10 @@ interface Props {
   nextResults: ResultsSet | null;
   setNextResults: (nextResults: ResultsSet) => void;
   selectedResult: MarkerIdAndInfo | null;
-  setSelectedResult: (selectedResult: MarkerIdAndInfo | null) => void;
+  setSelectedResult: (
+    selectedResult: MarkerIdAndInfo | null,
+    showDetails?: boolean,
```
jensren: Could you rename `showDetails` to `accordianView` and negate the boolean?

map/src/components/resultDetail.tsx
```diff
@@ -0,0 +1,247 @@
...
+}
+
+const Div = styled('div')`
+  overflow-y: auto;
+  max-height: 100%;
+  background: #fff;
+  padding: ${p => p.theme.spacingPx}px;
+  overflow-y: auto;
+  background: #fff;
+  box-shadow: 0px 4px 12px rgba(0, 0, 0, 0.15);
```
jensren, line 23: Do we have theme colours we can use for these instead of defining them?

map/src/components/resultDetail.tsx
```diff
@@ -0,0 +1,247 @@
...
+  border-bottom-right-radius: 4px;
+  cursor: auto;
+  pointer-events: initial;
+
+  > .disclaimer,
+  > .source {
+    background: rgba(0, 0, 0, 0.05);
+    border-radius: 3px;
+    padding: 10px 8px;
+    color: rgba(0, 0, 0, 0.5);
```
jensren, line 34: Same comment about the theme colours

map/src/components/resultDetail.tsx
```diff
@@ -0,0 +1,247 @@
...
+  }
+
+  > .name {
+    font-size: 1.5rem;
+    padding-bottom: ${p => p.theme.spacingPx / 2}px;
+  }
+
+  > .location {
+    color: ${p => p.theme.textColorLight};
+    font-size: 0.9rem;
```
jensren, line 59: Is there a reason this one is in `rem` and the other font sizes are in `px`? I feel it would be better to be consistent unless you have a specific reason to do it this way

map/src/components/resultDetail.tsx
```diff
@@ -0,0 +1,247 @@
...
+import styled from '../styling';
+import MarkerType from './marker-type';
+
+interface Props {
+  result: MarkerIdAndInfo;
+  lang: Language;
+  fullDetail?: boolean;
+}
+
+const Div = styled('div')`
```
jensren: Can we name this something more descriptive, e.g. `ResultDetailDiv`?

map/src/components/resultDetail.tsx
```diff
@@ -0,0 +1,247 @@
...
+          <li key={i}>{item}</li>
+        ))}
+      </ul>
+    </div>
+  );
+};
+
+const ResultDetail = ({ result, lang, fullDetail = false }: Props) => {
+  const selectedResultSource =
+    result?.info.source && SOURCES[result.info.source.name];
```
jensren: Do you need `result?`? It doesn't look like `result` is optional

map/src/components/resultDetail.tsx
```diff
@@ -0,0 +1,247 @@
...
+  const selectedResultSentenceType =
+    result?.info.type.type === 'mutual-aid-group' ||
+    result?.info.type.type === 'org'
```
jensren: Why mutual aid group and org?

map/src/components/resultDetail.tsx
```diff
@@ -0,0 +1,247 @@
...
+        </div>
+      )}
+      {result.info.contentBody && (
+        <div>
+          <div className="content">{result.info.contentBody}</div>
+        </div>
+      )}
+      {result.info.contact && (
+        <div>
+          {contactInfo(
```
jensren, line 226: If you want to write this as a `map`, you can use `Object.keys` to access all the keys as an array 😄 it's ok to leave this in my opinion since there are only 3, but just saying since I'm not sure if you're aware of `Object.keys`

map/src/components/resultDetail.tsx
pratikmdhr: @jensren will check the theme, i believe this styling code was already there

map/src/components/resultDetail.tsx
pratikmdhr: @jensren same here

map/src/components/resultDetail.tsx
pratikmdhr: same here

map/src/components/resultDetail.tsx
pratikmdhr: @jensren this code was there already too, I wasn't really sure about this either..

map/src/components/resultDetail.tsx
pratikmdhr: haha thats true

map/src/components/resultDetail.tsx
pratikmdhr: @jensren will change rem to px, seems like majority of the code is using px

map/src/components/resultDetail.tsx
pratikmdhr: will leave this one out for next PR, it seems like the white color is there in the theme with the name `background`, but noticed #fff is being used at most of the places

map/src/App.tsx
```diff
@@ -68,8 +70,14 @@ class App extends React.Component<Props, State> {
...
 
-  private setSelectedResult = (selectedResult: MarkerIdAndInfo | null) => {
-    this.setState({ selectedResult });
+  private setSelectedResult = (
+    selectedResult: MarkerIdAndInfo | null,
+    markerClicked = false,
+  ) => {
+    this.setState({
+      selectedResult,
+      mapMarkerClicked: markerClicked,
```
pratikmdhr, line 79: renamed this as `mapMarkerClicked` instead, because this state has a larger scope and needs to be accessed by the Map.tsx too.. 

Also the state for Accordion is handled inside Results.tsx, so i renamed the state to `expandAccordion` there for clarity (previously it was `expandResult`)

map/src/components/resultDetail.tsx
pratikmdhr: will leave this one out  for next PR too, it was part of old code

map/src/components/map.tsx
pratikmdhr: https://github.com/reach4help/reach4help/pull/1776#discussion_r851697991

map/src/App.tsx
pratikmdhr: https://github.com/reach4help/reach4help/pull/1776#discussion_r841234717

jensren (approved):
LGTM!

PR #1779, opened 2022-04-03
refactor: create new reusuable components for result and resultDetail

pratikmdhr:
## Description
Refactored `Results`  component and added two reusable components `Result` and `ResultDetail`.

## Motivation
To make `Results` component more clean, and managing state for an accordion becomes more simpler if there is a separate `Result` child component.

### DIFF
--- a/map/src/components/result.tsx
+++ b/map/src/components/result.tsx
@@ -1,0 +1,65 @@
+import React from 'react';
+import { MarkerIdAndInfo } from 'src/components/map';
+
+import styled from '../styling';
+import MarkerType from './marker-type';
+
+interface Props {
+  result: MarkerIdAndInfo | null;
+  index: number;
+  setSelectedResult: (selectedResult: MarkerIdAndInfo | null) => void;
+}
+
+const Div = styled('div')`
+  color: ${p => p.theme.textColor};
+  border-bottom: ${p => p.theme.borderLight};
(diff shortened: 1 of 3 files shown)


map/src/components/result.tsx
```diff
@@ -0,0 +1,65 @@
...
+import styled from '../styling';
+import MarkerType from './marker-type';
+
+interface Props {
+  result: MarkerIdAndInfo | null;
+  index: number;
+  setSelectedResult: (selectedResult: MarkerIdAndInfo | null) => void;
+}
+
+const Div = styled('div')`
```
pratikmdhr, line 13: @jensren I tried to apply styling to the class itself like in the [`Results` component](https://github.com/reach4help/reach4help/blob/a6809bdb23c3786a5ce75af38f4858a15ea1ed09/map/src/components/results.tsx#L129-L131)  , but not sure why it didn't work here, so I used the regular styled components for styling. Could you try it once and see if it works??

map/src/components/resultDetail.tsx
```diff
@@ -0,0 +1,240 @@
...
+
+import styled from '../styling';
+import MarkerType from './marker-type';
+
+interface Props {
+  result: MarkerIdAndInfo;
+  lang: Language;
+}
+
+const Div = styled('div')`
```
pratikmdhr, line 15: same here

